PR #24, opened 2022-11-28
feat: GO Feature Flag dotnet provider

thomaspoignant:
## Description
This PR adds a new provider for **GO Feature Flag**.


toddbaert:
Thanks, as always, @thomaspoignant/

@kinyoklion @benjiro please don't feel obligated to review this, but I've added you if you're interested. I'll be taking a look at it myself as well.

src/OpenFeature.Contrib.Providers.GOFeatureFlag/GoFeatureFlagProvider.cs
```diff
@@ -0,0 +1,325 @@
...
+        }
+
+        /// <summary>
+        ///     validateInputOptions is validating the different options provided when creating the provider.
+        /// </summary>
+        /// <param name="options">Options used while creating the provider</param>
+        /// <exception cref="InvalidOption">if no options are provided or we have a wrong configuration.</exception>
+        private void ValidateInputOptions(GoFeatureFlagProviderOptions options)
+        {
+            if (options == null) throw new InvalidOption("No options provided");
```
toddbaert, line 42: I'd suggest using `string.IsNullOrEmpty()`

src/OpenFeature.Contrib.Providers.GOFeatureFlag/GoFeatureFlagProvider.cs
```diff
@@ -0,0 +1,325 @@
...
+        {
+            try
+            {
+                var resp = await CallApi(flagKey, defaultValue, context);
+                return new ResolutionDetails<bool>(flagKey, bool.Parse(resp.value.ToString()), ErrorType.None,
+                    resp.reason, resp.variationType);
+            }
+            catch (FormatException e)
+            {
+                throw new TypeMismatchError("Flag value " + flagKey + " had unexpected type.", e);
```
toddbaert, line 97: Consider using [string interpolation](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/interpolated)

src/OpenFeature.Contrib.Providers.GOFeatureFlag/GoFeatureFlagProvider.cs
```diff
@@ -0,0 +1,325 @@
...
+                {
+                    var current = arrayEnumerator.Current;
+                    var convertedValue = ConvertValue(current);
+                    if (convertedValue != null) arr.Add(convertedValue);
+                }
+
+                return new Value(arr);
+            }
+
+            throw new InvalidCastException();
```
toddbaert, line 322: This seems like a good use case for `ErrorType.ParseError`.

thomaspoignant:
I don't know why GitHub has removed @benjiro and @kinyoklion from this one when I have re-request the approval from @toddbaert 😞

src/OpenFeature.Contrib.Providers.GOFeatureFlag/GoFeatureFlagProvider.cs
```diff
@@ -0,0 +1,325 @@
...
+
+            if (value.ValueKind == JsonValueKind.False || value.ValueKind == JsonValueKind.True)
+                return new Value(value.GetBoolean());
+
+            if (value.ValueKind == JsonValueKind.Number) return new Value(value.GetDouble());
+
+            if (value.ValueKind == JsonValueKind.Object)
+            {
+                var dict = new Dictionary<string, Value>();
+                var objEnumerator = value.EnumerateObject();
```
toddbaert, line 295: `JsonElement.ObjectEnumerator` implements [IDisposable](https://docs.microsoft.com/en-us/dotnet/api/system.text.json.jsonelement.objectenumerator.dispose?view=net-6.0#system-text-json-jsonelement-objectenumerator-dispose). I think you should be using it withing a `using` statement to make sure it's memory is released in a timely manner.

test/OpenFeature.Contrib.Providers.GOFeatureFlag.Test/GoFeatureFlagProviderTest.cs
```diff
@@ -0,0 +1,443 @@
...
+                new("beta")
+            }))
+            .Build();
+    }
+
+
+    [Fact]
+    private void getMetadata_validate_name()
+    {
+        var provider = new GoFeatureFlagProvider(new GoFeatureFlagProviderOptions
```
toddbaert, line 86: I think I'd prefer to see "integration" tests as well, where an OpenFeature client is created, but I understand that's more difficult and I think this is sufficent.

toddbaert:
I think https://github.com/open-feature/dotnet-sdk-contrib/pull/24#discussion_r1036435370 is my only concern. Unless you know for sure that there's no need to call `.Dispose()` on the Enumerator, I think we should, or we might have memory leaks or a lot of GC.

src/OpenFeature.Contrib.Providers.GOFeatureFlag/GoFeatureFlagProvider.cs
thomaspoignant: That's the 1st time I write C# I did not even knew that `using` was a thing 🤦 .
Thanks for the help I will have a look at it.

test/OpenFeature.Contrib.Providers.GOFeatureFlag.Test/GoFeatureFlagProviderTest.cs
thomaspoignant: I have changed the test to use the SDK and not the provider only.
It was actually pretty easy to do it.

toddbaert (approved):
Thanks @thomaspoignant ! I've approved!

If you could add just one more thing before merge: a small readme!

thomaspoignant:
> If you could add just one more thing before merge: a small readme!

@toddbaert I have added the README, I guess we are good to go.

PR #42, opened 2023-02-21
feat: unix socket support for flagd provider

### LINKED ISSUE #35: [flagd-provider] Implement Unix socket support
Implement a unix socket connection support for flagd, similar to the [Java](https://github.com/open-feature/java-sdk-contrib/blob/103fe68da6f1947d97a2f33363cff16b7dd50ec1/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java#L116-L122) and [JS](https://github.com/open-feature/js-sdk-contrib/blob/main/libs/providers/flagd/src/lib/service/grpc/service.ts#L99) implementations, using similar configuration paradigms.

Dependent on https://github.com/open-feature/dotnet-sdk-contrib/pull/31

bacherfl:
Closes #35 

As discussed in the issue, I have added net5.0 and above to the TargetFrameworks and used preprocessor directives to only make use of the unix socket support in compatible dotnet frameworks (net5.0 and above).

### DIFF
--- a/build/Common.prod.props
+++ b/build/Common.prod.props
@@ -6,7 +6,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net462</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net462;net5.0;net6.0;net7.0</TargetFrameworks>
     <RepositoryType>git</RepositoryType>
     <RepositoryUrl>https://github.com/open-feature/dotnet-sdk</RepositoryUrl>
     <Description>OpenFeature is an open standard for feature flag management, created to support a robust feature flag ecosystem using cloud native technologies. OpenFeature will provide a unified API and SDK, and a developer-first, cloud-native implementation, with extensibility for open source and commercial offerings.</Description>
--- a/src/OpenFeature.Contrib.Providers.Flagd/FlagdProvider.cs
+++ b/src/OpenFeature.Contrib.Providers.Flagd/FlagdProvider.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Linq;
 using System.Threading.Tasks;
+
 using Google.Protobuf.WellKnownTypes;
 using Grpc.Core;
 using Grpc.Net.Client;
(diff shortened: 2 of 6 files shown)


build/Common.prod.props
```diff
@@ -6,7 +6,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net462</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net462;net5.0;net6.0;net7.0</TargetFrameworks>
```
toddbaert, line 9: We might want to add `net7.0` to the matrix in the [linux](https://github.com/open-feature/dotnet-sdk-contrib/blob/main/.github/workflows/linux-ci.yml) and [windows](https://github.com/open-feature/dotnet-sdk-contrib/blob/main/.github/workflows/windows-ci.yml) ci workflows.

src/OpenFeature.Contrib.Providers.Flagd/FlagdProvider.cs
```diff
@@ -373,13 +386,39 @@ private static Value ConvertToPrimitiveValue(ProtoValue value)
...
         {
-#if NETSTANDARD2_0
-            return new Service.ServiceClient(GrpcChannel.ForAddress(url));
-#else
+            var useUnixSocket = url.ToString().StartsWith("unix://");
+#if NET462
+            if (useUnixSocket) {
+                // unix socket support is not available in this dotnet version
+                throw new Exception("unix sockets are not supported in this version.");
+            }
```
toddbaert, line 394: I _think_ this can be slightly simplified by using something like `if (!useUnixSocket) return...` here and moving the thrown exception to the very end outside of any compiler directives as a "catch-all". Then the exception can be thrown from only one place if no other `return` has occurred.

toddbaert:
I'd love to add some basic tests here but my (somewhat dated) .NET + Moq knowledge tells me it's not possible to test static methods or constructors with the tooling we're using.

toddbaert (approved):
Approved with minor comments/suggestions.

src/OpenFeature.Contrib.Providers.Flagd/FlagdProvider.cs
bacherfl: thanks for the suggestion, that definitely makes sense - I will give this and a shot and update the PR, if it works :)

bacherfl: you were right, using your suggested approach I was able to simplify the method a little bit 👍

toddbaert (approved):

PR #44, opened 2023-03-20
feat: added OTel hook

### LINKED ISSUE #28: Add OTel Hook
Now that [OTel 2529](https://github.com/open-telemetry/opentelemetry-specification/pull/2529) is merged, we should implement the Open Telemetry hook (currently exists as a stub).

Basically, this would entail creating a hook which:

- in the `after` stage: gets the current span and adds an appropriately named event (`feature_flag`), with the attributes (`feature_flag.key`, `feature_flag.provider_name`, and `feature_flag.variant`)
- in the `error` stage, records the exception

In pseudocode, this looks something like:

```
class OTelHook implements Hook {
  after(context, details) {
    var currentTrace = trace.getActiveSpan();
    currentTrace.addEvent("feature_flag", { "feature_flag.key": details.key, /*  other props */ } );
  }

  error(err) {
    var currentTrace = trace.getActiveSpan();
    currentTrace.recordException(err);
  }
}
```
See: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/feature-flags.md

bacherfl:
Closes #28 

This PR adds the implementation for the Otel hook


src/OpenFeature.Contrib.Hooks.Otel/OpenFeature.Contrib.Hooks.Otel.csproj
```diff
@@ -12,4 +12,9 @@
     <Authors>Todd Baert</Authors>
```
toddbaert, line 12: @bacherfl I think this is wrong :wink:

src/OpenFeature.Contrib.Hooks.Otel/OtelHook.cs
```diff
@@ -0,0 +1,64 @@
...
+            var span = Tracer.CurrentSpan;
+            if (span != null)
+            {
+                var attributes = new Dictionary<string, object>
+                {
+                    {"feature_flag.key", details.FlagKey},
+                    {"feature_flag.variant", details.Variant},
+                    {"feature_flag.provider_name", context.ProviderMetadata.Name}
+                };
+                span.AddEvent("feature_flag", new SpanAttributes(attributes));
```
toddbaert, line 36: If you feel like it's an improvement, you could use [string interpolation](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/interpolated) to extract `"feature_flag"` into a constant, but I think that's entirely up to your preference.

test/OpenFeature.Contrib.Hooks.Otel.Test/OtelHookTest.cs
```diff
@@ -0,0 +1,218 @@
...
+            var eventsEnum = rootSpan.Events.GetEnumerator();
+            eventsEnum.MoveNext();
+
+            ActivityEvent ev = (ActivityEvent)eventsEnum.Current;
+            Assert.Equal("feature_flag", ev.Name);
+
+            bool foundFlagKey = false;
+            bool foundFlagVariant = false;
+            bool foundFlagProviderName = false;
+            var tagsEnum = ev.Tags.GetEnumerator();
```
toddbaert, line 61: If you add `using System.Linq;` to the top of this file, you will find the `IEnumerable` interface (the type of `ev.Tags`) gains literally hundreds of convenience methods for iterating, filtering, mapping, reducing, etc. What you have works great, but you may prefer some of these.

toddbaert (approved):
Approved with optional comments.

toddbaert:
@beeme1mr or @thisthat if either of you could take a very brief glance at this (not from a C# perspective, but from an OTel concepts perspective) to make sure I'm not missing anything, that would be great.

thisthat (approved):
LGTM from a OTel PoV

test/OpenFeature.Contrib.Hooks.Otel.Test/OtelHookTest.cs
toddbaert: Nice changes!

toddbaert (approved):

PR #47, opened 2023-03-27
feat: implemented LRU caching for flagd provider 

### LINKED ISSUE #37: [flagd-provider] Implement standard caching, with LRU cache
Implement [standard flagd caching](https://github.com/open-feature/flagd/blob/main/docs/other_resources/caching.md). Use a LRU cache implementation such as [caffeine](https://github.com/ben-manes/caffeine) or similar. The cache requires a persistent connection to the flagd event stream. The stream should reconnect automatically if broken, and purge entries if change events are received or the connection fails.

See [JS](https://github.com/open-feature/js-sdk-contrib/blob/7401481c75416283c134184bc41f5b89dcb0ec49/libs/providers/flagd/src/lib/service/grpc/service.ts#L78) or [Java](https://github.com/open-feature/java-sdk-contrib/blob/main/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdCache.java) implementation for reference.

Depends on: https://github.com/open-feature/dotnet-sdk-contrib/pull/31

bacherfl:
Closes #37 

This PR implements standard caching functionality for the flagd provider. The configuration is done via the following env vars:

```
| Caching                      | FLAGD_CACHE                    | string  |           |     LRU       |
| Maximum cache size           | FLAGD_MAX_CACHE_SIZE           | number  | 10        |               |
| Maximum event stream retries | FLAGD_MAX_EVENT_STREAM_RETRIES | number  | 3         |               |
```

When caching is enabled, the flagd provider will establish an `EventStream` with the flagd server via grpc, and listen for `configuration_change` events. When such an event is received for a flag that has already been cached previously, that flag will be removed from the cache. When retrieving the value of a feature flag, the value will be cached if the `Reason` of the resolution is set to `STATIC`.

*Note:* The imported version of the OF dotnet-sdk is set to `0.5` - Since this version does not yet include the `configuration_change` event type, or the `STATIC` reason type for the resolutionDetails, it might be worth to consider bumping the version of the imported dotnet-sdk as a follow up task.

*Note:* Since the Event Stream is opened in a background task, I was not quite sure how to best verify in the Unit tests that things like the removal of a cached item were properly called by the provider. Coming from Golang, I was missing something like `require.Eventually(t, len(mockCache.DeleteCalls) == 1)`.  I tried to resemble that behaviour by creating an `AutoResetEvent` that gets set in the mock implementation of the method I would like to ensure it was called. Then, using `Assert.True(_autoResetEvent.WaitOne())` I waited for the event to be triggered. This solution does seem to work, but if there is a better approach to this in C# I appreciate every input from the reviewers of this PR

### DIFF
--- a/src/OpenFeature.Contrib.Providers.Flagd/Cache.cs
+++ b/src/OpenFeature.Contrib.Providers.Flagd/Cache.cs
@@ -1,8 +1,11 @@
 using System.Collections.Generic;
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("DynamicProxyGenAssembly2")]
 
 namespace OpenFeature.Contrib.Providers.Flagd
 {
-    interface ICache<TKey, TValue>
+    internal interface ICache<TKey, TValue>
     {
         void Add(TKey key, TValue value);
         TValue TryGet(TKey key);
--- a/src/OpenFeature.Contrib.Providers.Flagd/Cache.cs
+++ b/src/OpenFeature.Contrib.Providers.Flagd/Cache.cs
@@ -1,0 +1,111 @@
+using System.Collections.Generic;
+
+namespace OpenFeature.Contrib.Providers.Flagd
+{
+    interface ICache<TKey, TValue>
+    {
+        void Add(TKey key, TValue value);
+        TValue TryGet(TKey key);
+        void Delete(TKey key);
+    }
+    class LRUCache<TKey, TValue> : ICache<TKey, TValue> where TValue : class
+    {
+        private readonly int _capacity;
+        private readonly Dictionary<TKey, Node> _map;
+        private Node _head;
+        private Node _tail;
+
+        public LRUCache(int capacity)
+        {
+            _capacity = capacity;
+            _map = new Dictionary<TKey, Node>();
+        }
+
+        public TValue TryGet(TKey key)
+        {
+            System.Console.WriteLine("looking for " + key.ToString());
+            if (_map.TryGetValue(key, out Node node))
+            {
+                System.Console.WriteLine("found: " + key.ToString() + " with value: " + node.ToString());
+                MoveToFront(node);
+                return node.Value;
+            }
+            return default(TValue);
+        }
+
+        public void Add(TKey key, TValue value)
+        {
+            if (_map.TryGetValue(key, out Node node))
+            {
+                node.Value = value;
+                MoveToFront(node);
+            }
+            else
+            {
+                if (_map.Count == _capacity)
+                {
+                    _map.Remove(_tail.Key);
+                    RemoveTail();
+                }
+                node = new Node(key, value);
+                _map.Add(key, node);
+                AddToFront(node);
+            }
+        }
+
+        private void MoveToFront(Node node)
+        {
+            if (node == _head)
+                return;
+            node.Prev.Next = node.Next;
+            if (node == _tail)
+                _tail = node.Prev;
+            else
+                node.Next.Prev = node.Prev;
+            AddToFront(node);
+        }
+
+        private void AddToFront(Node node)
+        {
+            if (_head == null)
+            {
+                _head = node;
+                _tail = node;
+                return;
+            }
+            node.Next = _head;
+            _head.Prev = node;
+            _head = node;
+        }
+
+        private void RemoveTail()
+        {
+            _tail = _tail.Prev;
+            if (_tail != null)
+                _tail.Next = null;
+            else
+                _head = null;
+        }
+
+        public void Delete(TKey key)
+        {
+            throw new System.NotImplementedException();
+        }
+
+        private class Node
+        {
+            public TKey Key;
+            public TValue Value;
+            public Node Next;
+            public Node Prev;
+
+            public Node(TKey key, TValue value)
+            {
+                Key = key;
+                Value = value;
+            }
+        }
+    }
+
+}
+
--- a/src/OpenFeature.Contrib.Providers.Flagd/Cache.cs
+++ b/src/OpenFeature.Contrib.Providers.Flagd/Cache.cs
@@ -44,7 +44,7 @@ namespace OpenFeature.Contrib.Providers.Flagd
             {
                 _mtx.ReleaseMutex();
             }
-            
+
         }
 
         public void Add(TKey key, TValue value)
@@ -73,7 +73,7 @@ namespace OpenFeature.Contrib.Providers.Flagd
             {
                 _mtx.ReleaseMutex();
             }
-            
+
         }
 
         public void Delete(TKey key)
@@ -86,8 +86,8 @@ namespace OpenFeature.Contrib.Providers.Flagd
                     if (node == _head)
                     {
                         _head = node.Next;
-                    } 
-                    else 
+                    }
+                    else
                     {
                         node.Prev.Next = node.Next;
                     }
--- a/src/OpenFeature.Contrib.Providers.Flagd/Cache.cs
+++ b/src/OpenFeature.Contrib.Providers.Flagd/Cache.cs
@@ -30,9 +30,9 @@ namespace OpenFeature.Contrib.Providers.Flagd
 
         public TValue TryGet(TKey key)
         {
-            try
+            using (var mtx = new Mutex(ref _mtx))
             {
-                _mtx.WaitOne();
+                mtx.Lock();
                 if (_map.TryGetValue(key, out Node node))
                 {
                     MoveToFront(node);
@@ -40,18 +40,13 @@ namespace OpenFeature.Contrib.Providers.Flagd
                 }
                 return default(TValue);
             }
-            finally
-            {
-                _mtx.ReleaseMutex();
-            }
-
         }
 
         public void Add(TKey key, TValue value)
         {
-            try
+            using (var mtx = new Mutex(ref _mtx))
             {
-                _mtx.WaitOne();
+                mtx.Lock();
                 if (_map.TryGetValue(key, out Node node))
                 {
                     node.Value = value;
@@ -69,18 +64,13 @@ namespace OpenFeature.Contrib.Providers.Flagd
                     AddToFront(node);
                 }
             }
-            finally
-            {
-                _mtx.ReleaseMutex();
-            }
-
         }
 
         public void Delete(TKey key)
         {
-            try
+            using (var mtx = new Mutex(ref _mtx))
             {
-                _mtx.WaitOne();
+                mtx.Lock();
                 if (_map.TryGetValue(key, out Node node))
                 {
                     if (node == _head)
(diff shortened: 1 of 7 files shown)


src/OpenFeature.Contrib.Providers.Flagd/Cache.cs
```diff
@@ -0,0 +1,170 @@
...
+                {
+                    MoveToFront(node);
+                    return node.Value;
+                }
+                return default(TValue);
+            }
+            finally
+            {
+                _mtx.ReleaseMutex();
+            }
```
toddbaert, line 46: You may want to consider wrapping your mutex is an IDisposable interface, which would allow you to remove all these `finally` blocks. Here's a blog post explaining it: https://www.twilio.com/blog/c-8-making-use-of-using-declarations

src/OpenFeature.Contrib.Providers.Flagd/Cache.cs
```diff
@@ -0,0 +1,170 @@
...
+            {
+                _mtx.WaitOne();
+                if (_map.TryGetValue(key, out Node node))
+                {
+                    node.Value = value;
+                    MoveToFront(node);
+                }
+                else
+                {
+                    if (_map.Count == _capacity)
```
toddbaert, line 62: I understand that with the locking it should be impossible to have this condition, but it might be safer to make this a `>=` just in case.

src/OpenFeature.Contrib.Providers.Flagd/Cache.cs
```diff
@@ -0,0 +1,170 @@
...
+    internal interface ICache<TKey, TValue>
+    {
+        void Add(TKey key, TValue value);
+        TValue TryGet(TKey key);
+        void Delete(TKey key);
+        void Purge();
+    }
+    class LRUCache<TKey, TValue> : ICache<TKey, TValue> where TValue : class
+    {
+        private readonly int _capacity;
```
toddbaert, line 17: A c# int is a signed 32 bit, so max value here is `2,147,483,647`. If each of those entries is 10 bytes (probably a reasonable size for a flag-resolution) that's about ~200MB, which seems a bit small. I think perhaps this should be a long?

TBH I think that we might need the same change in Java. What do you think?

src/OpenFeature.Contrib.Providers.Flagd/FlagdProvider.cs
```diff
@@ -22,44 +22,44 @@ namespace OpenFeature.Contrib.Providers.Flagd
...
-        ///     FLAGD_PORT         - The port of the flagd server (default="8013")
-        ///     FLAGD_TLS          - Determines whether to use https or not (default="false")
-        ///     FLAGD_SOCKET_PATH - Path to the unix socket (default="")
+        ///     FLAGD_HOST                     - The host name of the flagd server (default="localhost")
+        ///     FLAGD_PORT                     - The port of the flagd server (default="8013")
+        ///     FLAGD_TLS                      - Determines whether to use https or not (default="false")
+        ///     FLAGD_SOCKET_PATH              - Path to the unix socket (default="")
+        ///     FLAGD_CACHE                    - Enable or disable the cache (default="false")
+        ///     FLAGD_MAX_CACHE_SIZE           - The maximum size of the cache (default="10")
+        ///     FLAGD_MAX_EVENT_STREAM_RETRIES - The maximum amount of retries for establishing the EventStream
```
toddbaert, line 45: The env stuff is important, but I think with all these additions, we _may_ want a constructor that also takes the config object directly.

Doesn't have to be part of this PR.

test/OpenFeature.Contrib.Providers.Flagd.Test/FlagdProviderTest.cs
```diff
@@ -305,5 +309,246 @@ public void TestResolveUnknownError()
...
+            var asyncStreamReader = new Mock<IAsyncStreamReader<EventStreamResponse>>();
+
+            var l = new List<EventStreamResponse>
+            {
+                new EventStreamResponse{
+                    Type = "provider_ready"
+                }
+            };
+
+            var enumerator = l.GetEnumerator();
```
toddbaert, line 342: Optional, but you can use `Linq` here.

test/OpenFeature.Contrib.Providers.Flagd.Test/FlagdProviderTest.cs
```diff
@@ -305,5 +309,246 @@ public void TestResolveUnknownError()
...
```
toddbaert, line 403: More optional `Linq`.

src/OpenFeature.Contrib.Providers.Flagd/FlagdProvider.cs
```diff
@@ -265,6 +321,94 @@ private FeatureProviderException GetOFException(Grpc.Core.RpcException e)
...
+                return;
+            }
+
+            try
+            {
+                if (data.Fields.TryGetValue("flags", out ProtoValue val))
+                {
+                    if (val.KindCase == ProtoValue.KindOneofCase.StructValue)
+                    {
+                        var changedFlags = val.StructValue.Fields.AsEnumerable().GetEnumerator();
```
toddbaert, line 371: More optional `Linq`.

src/OpenFeature.Contrib.Providers.Flagd/FlagdConfig.cs
```diff
@@ -0,0 +1,76 @@
...
+        private string _port;
+        private bool _useTLS;
+        private string _socketPath;
+        private bool _cache;
+        private int _maxCacheSize;
+        private int _maxEventStreamRetries;
+
+        internal FlagdConfig()
+        {
+            _host = Environment.GetEnvironmentVariable("FLAGD_HOST") ?? "localhost";
```
toddbaert, line 41: Up to you, but you can extract these as constants.

test/OpenFeature.Contrib.Providers.Flagd.Test/CacheTest.cs
```diff
@@ -0,0 +1,117 @@
+using Xunit;
+
+namespace OpenFeature.Contrib.Providers.Flagd.Test
+{
+    public class UnitTestLRUCache
```
toddbaert, line 5: You may want to add one very simple test/retrieve test. Of course, this case is indirectly tested by the other cases, but without that simple case it might be confusing if there's a regression in that simple funcitonaltiy.

test/OpenFeature.Contrib.Providers.Flagd.Test/FlagdProviderTest.cs
```diff
@@ -305,5 +309,246 @@ public void TestResolveUnknownError()
...
+            var val = flagdProvider.ResolveBooleanValue("my-key", false, null);
+            Assert.True(val.Result.Value);
+
+            // set firstCall to true to make the mock EventStream return a configuration_change event
+            firstCall = false;
+
+            val = flagdProvider.ResolveBooleanValue("my-key", false, null);
+            Assert.True(val.Result.Value);
+
+            Assert.True(_autoResetEvent.WaitOne());
```
toddbaert, line 548: Add a timeout here (a param to `WaitOne`) to prevent tests for hanging.

toddbaert:
> Note: Since the Event Stream is opened in a background task, I was not quite sure how to best verify in the Unit tests that things like the removal of a cached item were properly called by the provider. Coming from Golang, I was missing something like require.Eventually(t, len(mockCache.DeleteCalls) == 1). I tried to resemble that behaviour by creating an AutoResetEvent that gets set in the mock implementation of the method I would like to ensure it was called. Then, using Assert.True(_autoResetEvent.WaitOne()) I waited for the event to be triggered. This solution does seem to work, but if there is a better approach to this in C# I appreciate every input from the reviewers of this PR

I think your testing strategy is fine. I find the tests around caching about as readable as possible given the complexity. They look similar to the ones in other contribs.

I think my only request with them is that you add timeouts to every `WaitOne()` to prevent the tests from hanging forever in cases where they break because the awaited function is not called. I purposely broke some tests locally and it seems like Xunit happily runs until I manually kill the process :joy:

toddbaert (changes requested):
Hey @bacherfl , IMO this all looks really great. I love your cache impl. I've left mostly optional/suggested changes except [this one](https://github.com/open-feature/dotnet-sdk-contrib/pull/47#issuecomment-1489167391), which seems pretty small but will prevent CI/tests from hanging if people break something when making changes.

bacherfl:
> Hey @bacherfl , IMO this all looks really great. I love your cache impl. I've left mostly optional/suggested changes except [this one](https://github.com/open-feature/dotnet-sdk-contrib/pull/47#issuecomment-1489167391), which seems pretty small but will prevent CI/tests from hanging if people break something when making changes.

Thanks @toddbaert for your review and suggestions - I agree with all of them, and will try to include them into this PR

src/OpenFeature.Contrib.Providers.Flagd/Cache.cs
bacherfl: that's a great suggestion, thanks! I was not too happy about the `finally` blocks either, especially being used to Go's `defer`. The approach with the `IDisposable` interface will be much closer to what I'm used to with Go, so I'll be happy to look into that :)

src/OpenFeature.Contrib.Providers.Flagd/FlagdProvider.cs
bacherfl: Agree, that would certainly make sense. If it's ok, I would do this in a follow up PR

src/OpenFeature.Contrib.Providers.Flagd/FlagdProvider.cs
```diff
@@ -231,12 +273,26 @@ public override async Task<ResolutionDetails<Value>> ResolveStructureValue(strin
             }, context);
         }
 
-        private async Task<ResolutionDetails<T>> ResolveValue<T>(Func<Struct, Task<ResolutionDetails<T>>> resolveDelegate, EvaluationContext context = null)
+        private async Task<ResolutionDetails<Value>> ResolveValue<T>(string flagKey, Func<Struct, Task<ResolutionDetails<Value>>> resolveDelegate, EvaluationContext context = null)
```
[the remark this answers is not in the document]
bacherfl: thanks for the suggestion, I have updated the function accordingly

bacherfl:
alright, I have submitted all changes/suggestions, so the PR should be ready for another review :)

src/OpenFeature.Contrib.Providers.Flagd/FlagdProvider.cs
toddbaert: yep!

toddbaert (approved):
lgtm!

PR #57, opened 2023-04-04
feat: introduce FlagdProvider constructor accepting FlagdConfig as parameter

### LINKED ISSUE #56: [flagd-provider] add constructor with configuration object
The flagd provider currently supports getting config from env-vars, but we should also add the ability to configure it with a simple config object, as discussed [here](https://github.com/open-feature/dotnet-sdk-contrib/pull/47#discussion_r1152321671) and [here](https://github.com/open-feature/dotnet-sdk-contrib/pull/48#discussion_r1153689674) with @bacherfl and @odubajDT .

Configurations provided in this options object should override values from the env vars, as is the case in the java/js equivalent.

odubajDT:
## This PR

- introduces FlagdPorvider constructor accepting FlagdConfig as parameter
- exposes FlagdConfig as public class
- adds unit test for constructor

### Related Issues

Fixes #56

### DIFF
--- a/src/OpenFeature.Contrib.Providers.Flagd/FlagdConfig.cs
+++ b/src/OpenFeature.Contrib.Providers.Flagd/FlagdConfig.cs
@@ -3,7 +3,10 @@ using System;
 namespace OpenFeature.Contrib.Providers.Flagd
 
 {
-    internal class FlagdConfig
+    /// <summary>
+    ///     FlagdConfig is the configuration object for flagD.
+    /// </summary>
+    public class FlagdConfig
     {
         internal const string EnvVarHost = "FLAGD_HOST";
         internal const string EnvVarPort = "FLAGD_PORT";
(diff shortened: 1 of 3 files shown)


toddbaert:
This looks good!

In other providers, we make sure that configs passed in the constructor "win" over env vars. Looking at how the config object works I think that would be the case here, which is good, but I don't see tests for that. Do you think you could add some assertions for this?

odubajDT:
> This looks good!
> 
> In other providers, we make sure that configs passed in the constructor "win" over env vars. Looking at how the config object works I think that would be the case here, which is good, but I don't see tests for that. Do you think you could add some assertions for this?

Surely, I can work on it :)

odubajDT:
@toddbaert enhanced the tests to check that the env variables are ignored, when using the FlagdProvider constructor accepting the FlagdConfig as a parameter :)

bacherfl (approved):

toddbaert (approved):

PR #72, opened 2023-08-04
fix: NET462 requires TLS for GRPC to work

Wolftousen:
## This PR
Currently, this provider is listed on NuGet.org as being compatible with .Net Framework 4.6.2+.  While this does compile and link properly, it does not function at all with that.  .Net Framework 4.6.2+ requires TLS to be used with GRPC or else GRPC flat out doesn't work, specified here in their [documentation](https://learn.microsoft.com/en-us/aspnet/core/grpc/netstandard?view=aspnetcore-7.0#net-framework).

The code here verifies that the cert provided matches the server cert and ignores self signed errors for .Net Framework 4.6.2 through .Net Framework 4.8.1.

### Related Issues
#73 

### Notes
I also ran into a compiler error on line 336 with `new Empty()` not being the right type and fixed it.  My limited testing shows it had no adverse effects.

### How to test

Create a self signed server certificate and launch the Flagd docker image with it as well as some of the sample flag data.  Then launch a dummy client that uses the provider and is also using the server certificate and TLS env vars setup, then just evaluate one of the flags.

### DIFF
--- a/src/OpenFeature.Contrib.Providers.Flagd/FlagdProvider.cs
+++ b/src/OpenFeature.Contrib.Providers.Flagd/FlagdProvider.cs
@@ -559,7 +559,7 @@ namespace OpenFeature.Contrib.Providers.Flagd
                         };
 #elif NET462_OR_GREATER
                         handler.ServerCertificateValidationCallback = (message, cert, chain, errors) => {
-                            if ((errors & SslPolicyErrors.None) > 0) { return true; }
+                            if (errors == SslPolicyErrors.None) { return true; }
 
                             chain.ChainPolicy.VerificationFlags = X509VerificationFlags.AllowUnknownCertificateAuthority;
 
--- a/src/OpenFeature.Contrib.Providers.Flagd/FlagdProvider.cs
+++ b/src/OpenFeature.Contrib.Providers.Flagd/FlagdProvider.cs
@@ -333,7 +333,7 @@ namespace OpenFeature.Contrib.Providers.Flagd
         {
             while (_eventStreamRetries < _config.MaxEventStreamRetries)
             {
-                var call = _client.EventStream(new EventStreamRequest());
+                var call = _client.EventStream(new Empty());
                 try
                 {
                     // Read the response stream asynchronously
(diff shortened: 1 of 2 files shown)


toddbaert:
Thanks @Wolftousen . I'll take a look at this later today. In the meantime, there's some errors in the CI. Specifically an issue with resolving an import (`EventStreamRequest`) and some lint errors.

src/OpenFeature.Contrib.Providers.Flagd/FlagdProvider.cs
```diff
@@ -539,30 +540,49 @@ private Service.ServiceClient BuildClientForPlatform(Uri url)
...
                             chain.ChainPolicy.TrustMode = X509ChainTrustMode.CustomRootTrust;
                             chain.ChainPolicy.CustomTrustStore.Add(certificate);
                             return chain.Build(cert);
                         };
-                    } else {
-                        throw new ArgumentException("Specified certificate cannot be found.");
-                    }
+#elif NET462_OR_GREATER
+                        handler.ServerCertificateValidationCallback = (message, cert, chain, errors) => {
+                            if ((errors & SslPolicyErrors.None) > 0) { return true; }
```
toddbaert, line 562: `errors` is of type `SslPolicyErrors`, correct? So this is essentially a ANDing some value (0,1,2,4) with 0, so won't this never return true?

Maybe I'm missing something here, but I think this return will never execute.

toddbaert:
Largely this looks good. I have one question about [this](https://github.com/open-feature/dotnet-sdk-contrib/pull/72#discussion_r1284737471), and one additional request if you'd be so kind... could you mention something about this in the readme (the requirement that .NET Framework requires TLS, etc)?

src/OpenFeature.Contrib.Providers.Flagd/FlagdProvider.cs
Wolftousen: You are correct, had that fix in a stash I forgot about.

Wolftousen:
@toddbaert updated the readme to include a note about the limitation and a link to the Microsoft documentation

bacherfl (approved):
Thank you for that contribution! This looks good to me

beeme1mr:
Hey @Wolftousen, Todd is out until Friday but I don't think we need to wait. Your change looks good to me and @bacherfl has already approved. Thanks for your effort on this. 🥳

beeme1mr (approved):

PR #83, opened 2023-09-12
fix: Setup config correct when passing a Uri (fixes #71)

### LINKED ISSUE #71: [flagd-provider] FlagdProvider constructor which takes only Uri will throw object null reference
FlagdProvider(Uri url) does not initialize _config object hence using this constructor will leave _config null and throw exception.

bjoneill:
## This PR

Provider can now be created with a `Uri`. Anything related to the URI is no longer read from environment variables, everything else is.

### Related Issues
Fixes #71 

### Notes
My unix socket knowledge is pretty slim, unit test in place to cover the basics and appears to do as expected.

### How to test
N/A

### DIFF
--- a/test/OpenFeature.Contrib.Providers.Flagd.Test/FlagdConfigTest.cs
+++ b/test/OpenFeature.Contrib.Providers.Flagd.Test/FlagdConfigTest.cs
@@ -90,7 +90,7 @@ namespace OpenFeature.Contrib.Providers.Flagd.Test
             Assert.False(config.CacheEnabled);
             Assert.Equal(new System.Uri("http://domain:8123"), config.GetUri());
         }
-        
+
         [Fact]
         public void TestFlagdConfigFromUriHttps()
         {
@@ -100,7 +100,7 @@ namespace OpenFeature.Contrib.Providers.Flagd.Test
             Assert.False(config.CacheEnabled);
             Assert.Equal(new System.Uri("https://domain:8123"), config.GetUri());
         }
-        
+
         [Fact]
         public void TestFlagdConfigFromUriUnix()
         {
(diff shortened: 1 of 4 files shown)


test/OpenFeature.Contrib.Providers.Flagd.Test/FlagdConfigTest.cs
```diff
@@ -80,6 +81,80 @@ public void TestFlagdConfigSetCertificatePath()
...
+        }
+
+        [Fact]
+        public void TestFlagdConfigFromUriUnix()
+        {
+            CleanEnvVars();
+            var config = new FlagdConfig(new Uri("unix:///var/run/tmp.sock/"));
+
+            Assert.False(config.CacheEnabled);
+            Assert.Equal(new System.Uri("unix:///var/run/tmp.sock/"), config.GetUri());
```
toddbaert, line 111: nit: I think every use of `System.` in this file is un-needed.

test/OpenFeature.Contrib.Providers.Flagd.Test/FlagdConfigTest.cs
```diff
@@ -80,6 +81,80 @@ public void TestFlagdConfigSetCertificatePath()
...
+
+            Assert.False(config.CacheEnabled);
+            Assert.Equal(new System.Uri("https://domain:8123"), config.GetUri());
+        }
+
+        [Fact]
+        public void TestFlagdConfigFromUriUnix()
+        {
+            CleanEnvVars();
+            var config = new FlagdConfig(new Uri("unix:///var/run/tmp.sock/"));
```
toddbaert, line 108: nit: another minor one (and maybe I'm missing something :sweat_smile:) but the trailing `/` here makes the socket look like a dir (which is slightly odd/confusing). I know it's just test data, so it's not a big deal.

test/OpenFeature.Contrib.Providers.Flagd.Test/FlagdConfigTest.cs
bjoneill: Removed :)

test/OpenFeature.Contrib.Providers.Flagd.Test/FlagdConfigTest.cs
bjoneill: Tidied this up :)

PR #114, opened 2023-12-14
feat: Add Metrics Hook

### LINKED ISSUE #113: Create a new `MetricsHook`
See [parent](https://github.com/open-feature/dotnet-sdk-contrib/issues/70).

askpt:
## This PR

- This PR adds the metrics to the Feature Flag evaluation hook.

### Related Issues

Fixes #113


toddbaert:
@thisthat @beeme1mr adding you guys because you have a bit more OTel experience than me.

src/OpenFeature.Contrib.Hooks.Otel/MetricsHook.cs
```diff
@@ -0,0 +1,134 @@
...
+    /// Represents a hook for capturing metrics related to flag evaluations.
+    /// The meter name is "OpenFeature.Contrib.Hooks.Otel".
+    /// </summary>
+    public class MetricsHook : Hook
+    {
+        private static readonly AssemblyName AssemblyName = typeof(MetricsHook).Assembly.GetName();
+        private static readonly string InstrumentationName = AssemblyName.Name;
+        private static readonly string InstrumentationVersion = AssemblyName.Version.ToString();
+
+        private readonly UpDownCounter<double> _evaluationActiveUpDownCounter;
```
thisthat, line 21: since the values will always be natural numbers, I suggest using `long`
```suggestion
        private readonly UpDownCounter<long> _evaluationActiveUpDownCounter;
```

src/OpenFeature.Contrib.Hooks.Otel/MetricsConstants.cs
```diff
@@ -0,0 +1,23 @@
...
+        internal const string RequestsTotalName = "feature_flag.evaluation_requests_total";
+        internal const string SuccessTotalName = "feature_flag.evaluation_success_total";
+        internal const string ErrorTotalName = "feature_flag.evaluation_error_total";
+
+        internal const string ActiveDescription = "active flag evaluations counter";
+        internal const string RequestsDescription = "feature flag evaluation request counter";
+        internal const string SuccessDescription = "feature flag evaluation success counter";
+        internal const string ErrorDescription = "feature flag evaluation error counter";
+
+        internal const string KeyAttr = "key";
```
thisthat, line 17: We should look into why this pkg doesn't provide the attributes for FF: https://github.com/open-telemetry/opentelemetry-dotnet/tree/af83eb1043a34f82ba3bfe20c8b674ac295cfb37/src/OpenTelemetry.SemanticConventions

thisthat (commented):
LGTM, I only left a minor suggestion :) 
I don't know enough of C# to feel confident approving the PR tho, I'll delegate someone else for this :)

src/OpenFeature.Contrib.Hooks.Otel/MetricsHook.cs
thisthat: Yes please 👍

askpt: Solved with https://github.com/open-feature/dotnet-sdk-contrib/pull/114/commits/f8e52e6d3631b87b936b8dc56534ec41d189477f

toddbaert:
@beeme1mr is there a spec for these metrics somewhere? I know we have the trace semantics covered in OTel now - but I don't think there's anything specifically about metrics (correct me if I'm wrong)

> LGTM, I only left a minor suggestion :) I don't know enough of C# to feel confident approving the PR tho, I'll delegate someone else for this :)

Yes, thanks @thisthat . I will review from a C# perspective, just glad to have you vet the OTel-specific stuff.

src/OpenFeature.Contrib.Hooks.Otel/MetricsConstants.cs
toddbaert: Not sure if they expect PRs for this or not, but we could open one.

test/OpenFeature.Contrib.Hooks.Otel.Test/MetricsHookTest.cs
```diff
@@ -0,0 +1,163 @@
...
+    {
+        [Fact]
+        public void After_Test()
+        {
+            // Arrange metrics collector
+            var exportedItems = new List<Metric>();
+            Sdk.CreateMeterProviderBuilder()
+                .AddMeter("*")
+                .ConfigureResource(r => r.AddService("openfeature"))
+                .AddInMemoryExporter(exportedItems, option => option.PeriodicExportingMetricReaderOptions = new PeriodicExportingMetricReaderOptions { ExportIntervalMilliseconds = 100 })
```
toddbaert, line 23: I'm glad you used the in-memory exporter for the tests. This is nicer than mocking everything IMO, since it means we don't have to make as many assumptions. I did the same thing in the JS impl.

src/OpenFeature.Contrib.Hooks.Otel/MetricsConstants.cs
askpt: It seems the library has not "updated" since 2022: https://www.nuget.org/packages/OpenTelemetry.SemanticConventions/1.0.0-rc9.9

Might need some help to reopen engagement there

test/OpenFeature.Contrib.Hooks.Otel.Test/MetricsHookTest.cs
askpt: There was a small catch with that. Since we need to wait for the exporter, I needed to tune the periodic exporting to be more frequent and allow some buffer time for the metrics to be exported.

If this library was in net8, I would probably use the `IMeterFactory`: https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.metrics.imeterfactory?view=net-8.0

test/OpenFeature.Contrib.Hooks.Otel.Test/MetricsHookTest.cs
```diff
@@ -0,0 +1,163 @@
...
+            Assert.True(hookTask.IsCompleted);
+
+            // Assert metrics
+            Assert.NotEmpty(exportedItems);
+
+            // check if the metric is present in the exported items
+            var metric = exportedItems.FirstOrDefault(m => m.Name == metricName);
+            Assert.NotNull(metric);
+
+            var noOtherMetric = exportedItems.All(m => m.Name == metricName);
```
toddbaert, line 47: Could you enhance / add some some assertions here about the attributes [here](https://github.com/open-feature/dotnet-sdk-contrib/pull/114/files#diff-39cf92a95324d98f7eae3aac643b31749e76c89e4cd6223b6ed39c40381050daR78-R81)?

test/OpenFeature.Contrib.Hooks.Otel.Test/MetricsHookTest.cs
```diff
@@ -0,0 +1,163 @@
...
```
toddbaert, line 83: Same [as above](https://github.com/open-feature/dotnet-sdk-contrib/pull/114/files#r1428084208), I think this can be added for each stage so we don't have attribute regressions.

src/OpenFeature.Contrib.Hooks.Otel/README.md
```diff
@@ -64,6 +65,54 @@ In case something went wrong during a feature flag evaluation, you will see an e
 
 ![](./assets/otlp-error.png)
 
+## Usage - Metrics
+
+For this hook to function correctly a global `MeterProvider` must be set, an example of how to do this can be found below.
+
```
toddbaert, line 71: Could you copy the table [here](https://github.com/open-feature/java-sdk-contrib/tree/main/hooks/open-telemetry#metricshook) and include it in this readme? I think we should actually add a section to the OpenFeature spec about this, but for now it's helpful to have them in the README I think.

test/OpenFeature.Contrib.Hooks.Otel.Test/MetricsHookTest.cs
toddbaert: > There was a small catch with that. Since we need to wait for the exporter, I needed to tune the periodic exporting to be more frequent and allow some buffer time for the metrics to be exported.

I saw that. I think it's fine. The tests are still quick and I think the value in having an real exporter in the tests is worth a 150ms of waiting.

toddbaert (changes requested):
Really good work, much appreciated. I think my only concern is a lack of assertions on the attributes (if I missed that, please point me in the right direction).

askpt:
> Really good work, much appreciated. I think my only concern is a[ lack of assertions](https://github.com/open-feature/dotnet-sdk-contrib/pull/114#discussion_r1428084208) on the attributes (if I missed that, please point me in the right direction).

@toddbaert: after inspecting the Java library I noticed a few differences that I want to correct:

- [ ] Units are wrong. All are long and it should be another unit.
- [ ] Doesn't support at the moment what they called `Custom dimensions`.
- [ ] Needs assertions at the tag list level. I am currently investigating this since I cannot find an easy way to assert the tag list.

toddbaert:
> * [ ]  Doesn't support at the moment what they called `Custom dimensions`.

Doesn't support at the moment what they called Custom dimensions.

I noticed this, and it didn't immediately concern me because it's something that can be added later without breakage.

If you want, you can add it in this PR though.

> * [ ] Needs assertions at the tag list level. I am currently investigating this since I cannot find an easy way to assert the tag list.

Here's how it was implemented in JS: https://github.com/open-feature/js-sdk-contrib/blob/1db63dca9a392c14f1574ec5ea8bda53f44123cd/libs/hooks/open-telemetry/src/lib/metrics/metrics-hook.spec.ts#L103-L112

.github/component_owners.yml
```diff
@@ -4,6 +4,7 @@ components:
   src/OpenFeature.Contrib.Hooks.Otel:
     - bacherfl
     - toddbaert
+    - askpt
```
toddbaert, line 7: @askpt I hope you don't mind I've added you here. This will add you to PRs for this component automatically, [once you're an org member](https://github.com/open-feature/community/pull/309).

askpt:
> > * [ ]   Doesn't support at the moment what they called `Custom dimensions`.
> 
> I noticed this, and it didn't immediately concern me because it's something that can be added later without breakage.
> 
> If you want, you can add it in this PR though. Otherwise I'll simply create a new issue.

Please create a new issue. This PR is already big enough and it's not a breaking change as you mentioned.

> > * [ ]  Needs assertions at the tag list level. I am currently investigating this since I cannot find an easy way to assert the tag list.
> 
> Here's how it was implemented in JS: https://github.com/open-feature/js-sdk-contrib/blob/1db63dca9a392c14f1574ec5ea8bda53f44123cd/libs/hooks/open-telemetry/src/lib/metrics/metrics-hook.spec.ts#L103-L112, if that's at all helpful!

I am trying to find a way to do it but it doesn't seem possible. The counter tags are private in the meter class `tagsToMetricPointIndexDictionary`. This is all I can find:
Screenshot 2023-12-18 at 18 03 10

toddbaert:
> > > * [ ]    Doesn't support at the moment what they called `Custom dimensions`.
> > 
> > 
> > I noticed this, and it didn't immediately concern me because it's something that can be added later without breakage.
> > If you want, you can add it in this PR though. Otherwise I'll simply create a new issue.
> 
> Please create a new issue. This PR is already big enough and it's not a breaking change as you mentioned.

[Done](https://github.com/open-feature/dotnet-sdk-contrib/issues/118).

> > > * [ ]   Needs assertions at the tag list level. I am currently investigating this since I cannot find an easy way to assert the tag list.
> > 
> > 
> > Here's how it was implemented in JS: https://github.com/open-feature/js-sdk-contrib/blob/1db63dca9a392c14f1574ec5ea8bda53f44123cd/libs/hooks/open-telemetry/src/lib/metrics/metrics-hook.spec.ts#L103-L112, if that's at all helpful!
> 
> I am trying to find a way to do it but it doesn't seem possible. The counter tags are private in the meter class `tagsToMetricPointIndexDictionary`. This is all I can find: Screenshot 2023-12-18 at 18 03 10

OK in that case I'm tempted to call the testing good enough, given the limitations.

toddbaert (approved):
Approved in light of [this comment](https://github.com/open-feature/dotnet-sdk-contrib/pull/114#issuecomment-1861354234).

toddbaert:
I'll merge this EOD unless I hear objections.

PR #116, opened 2023-12-15
refactor: Replace OtelHook with TracingHook. Deprecating OtelHook

### LINKED ISSUE #112: Rename `OtelHook` into `TracingHook`
See [parent](https://github.com/open-feature/dotnet-sdk-contrib/issues/70).

askpt:
## This PR

- This PR renames the `OtelHook` into `TracingHook`, keeping the `OtelHook` for the clients.
- Deprecating `OtelHook` that should be removed in the following version.

### Related Issues

Fixes #112


src/OpenFeature.Contrib.Hooks.Otel/TracingHook.cs
```diff
@@ -11,7 +9,7 @@ namespace OpenFeature.Contrib.Hooks.Otel
     /// <summary>
     /// Stub.
     /// </summary>
-    public class OtelHook : Hook
+    public class TracingHook : Hook
```
askpt, line 12: I have been thinking if it would make sense to duplicate the `OtelHook` class that uses the `TracingHook`. My idea is to add the `Obsolete` attribute to the `OtelHook` and remove it later.

toddbaert: If you mean as just a ways to warn people of the breaking change, that's fine with me!

I guess the easiest way might be to extend the new `TracingHook`, but add no functions and then mark that deprecated?

 I would support that, then we could make this a non-breaking feat.

askpt: @toddbaert Added my suggestion. If you don't see any benefit, I can remove the last commit and merge as it was.

src/OpenFeature.Contrib.Hooks.Otel/OtelHook.cs
```diff
@@ -1,18 +1,29 @@
...
     /// Stub.
     /// </summary>
+    [ExcludeFromCodeCoverage]
+    [Obsolete("This class is obsolete and will be removed in a future version. Please use TracingHook instead.")]
     public class OtelHook : Hook
     {
+        private readonly TracingHook _tracingHook;
+
+        /// <summary>
+        /// Stub.
```
toddbaert, line 20: Can you remove this?

toddbaert (approved):
LGTM. Please just update the title to be consistent with this, and remove the `!`.

src/OpenFeature.Contrib.Hooks.Otel/OtelHook.cs
askpt: Fixed with https://github.com/open-feature/dotnet-sdk-contrib/pull/116/commits/00b17642ec7d4f8004ceeab48719de780544eabf

PR #119, opened 2023-12-21
feat: Add ConfigCat provider

luizbon:
## This PR
- Adding ConfigCat provider integration

### How to test
Have a ConfigCat free account, create Product/Config/Keys and use as any other OpenFeature provider.


src/OpenFeature.Contrib.ConfigCat/ConfigCatProvider.cs
```diff
@@ -0,0 +1,99 @@
...
+            var user = context?.BuildUser();
+            var result = await Client.GetValueDetailsAsync(flagKey, defaultValue, user);
+            return string.IsNullOrEmpty(result.ErrorMessage)
+                ? new ResolutionDetails<T>(flagKey, result.Value, variant: result.VariationId)
+                : new ResolutionDetails<T>(flagKey, defaultValue, ParseErrorType(result.ErrorMessage), errorMessage: result.ErrorMessage);
+        }
+
+        private static ErrorType ParseErrorType(string errorMessage)
+        {
+            if(errorMessage.Contains("Config JSON is not present when evaluating setting"))
```
laliconfigcat, line 80: As I see this check is unnecessary as the next .Contains() will do the same.

src/OpenFeature.Contrib.ConfigCat/ConfigCatProvider.cs
```diff
@@ -0,0 +1,99 @@
...
+        private async Task<ResolutionDetails<T>> ProcessFlag<T>(string flagKey, EvaluationContext context, T defaultValue)
+        {
+            var user = context?.BuildUser();
+            var result = await Client.GetValueDetailsAsync(flagKey, defaultValue, user);
+            return string.IsNullOrEmpty(result.ErrorMessage)
+                ? new ResolutionDetails<T>(flagKey, result.Value, variant: result.VariationId)
+                : new ResolutionDetails<T>(flagKey, defaultValue, ParseErrorType(result.ErrorMessage), errorMessage: result.ErrorMessage);
+        }
+
+        private static ErrorType ParseErrorType(string errorMessage)
```
laliconfigcat, line 78: As the error messages can change much easier, I'd go with checking for the error codes and levels:
https://github.com/configcat/.net-sdk/blob/master/src/ConfigCatClient/Logging/LogMessages.cs#L10
The format looks like this: https://github.com/configcat/.net-sdk/blob/master/src/ConfigCat.Client.Tests/data/evaluationlog/simple_value/int_setting.txt#L1

So e.g. instead of "Config JSON is not present", I'd check for "ERROR [1000]"

src/OpenFeature.Contrib.ConfigCat/ConfigCatProvider.cs
```diff
@@ -0,0 +1,99 @@
...
+            var result = await ProcessFlag(flagKey, context, stringDefaultValue);
+            var returnValue = result.Value == null ? defaultValue : new Value(result.Value);
+            return new ResolutionDetails<Value>(flagKey, returnValue, variant: result.Variant, errorMessage: result.ErrorMessage);
+        }
+
+        private async Task<ResolutionDetails<T>> ProcessFlag<T>(string flagKey, EvaluationContext context, T defaultValue)
+        {
+            var user = context?.BuildUser();
+            var result = await Client.GetValueDetailsAsync(flagKey, defaultValue, user);
+            return string.IsNullOrEmpty(result.ErrorMessage)
```
laliconfigcat, line 73: As the GetValueDetailsAsync handling everything  - it returns the defaultValue if needed, I'd not override this behaviour in any cases with giving back the defaultValue if the errorMessage is not empty.

I think this return can be simplified to 
`return new ResolutionDetails<T>(flagKey, result.Value, !string.IsNullOrEmpty(result.ErrorMessage) ? ParseErrorType(result.ErrorMessage) : null, variant: result.VariationId, errorMessage: result.ErrorMessage);`

src/OpenFeature.Contrib.ConfigCat/UserBuilder.cs
```diff
@@ -0,0 +1,52 @@
...
+    internal static class UserBuilder
+    {
+        internal static User BuildUser(this EvaluationContext context)
+        {
+            if(context == null)
+            {
+                return null;
+            }
+
+            var user = new User(Guid.NewGuid().ToString());
```
laliconfigcat, line 18: I'd only initialize the new User with a GUID if you could not parse the "id" or "identifier" from the context.

laliconfigcat (changes requested):
Hello @luizbon ,

Thank you very much for your contribution! We really appreciate it at ConfigCat.
I left some minor comments, but I think it looks good.

Have a nice day!

toddbaert:
I'll give this a review tomorrow! Thanks @luizbon

src/OpenFeature.Contrib.ConfigCat/ConfigCatProvider.cs
luizbon: oops... 🤣

src/OpenFeature.Contrib.ConfigCat/ConfigCatProvider.cs
luizbon: Yeah, I would love to use the error codes, but unfortunately the `ErrorMessage` does not contain it.
I also checked the exception object, and it does not contain the error code.
This code is only used in the Logger.
image

src/OpenFeature.Contrib.ConfigCat/ConfigCatProvider.cs
luizbon: Good suggestion.

src/OpenFeature.Contrib.ConfigCat/UserBuilder.cs
luizbon: Fair enough

src/OpenFeature.Contrib.ConfigCat/ConfigCatProvider.cs
laliconfigcat: OMG, I'm really sorry, I totally thought that we are giving back the error code too. 
I'll talk about this with the guys. Totally makes sense to give back the error code too.

luizbon:
Hi @laliconfigcat, I have addressed the points you mentioned. Thanks for the feedback.
There's only one pending regarding the error code, though.

laliconfigcat (approved):
Really good job! Thank you again for your efforts and for the quick changes.

src/OpenFeature.Contrib.ConfigCat/ConfigCatProvider.cs
toddbaert: I guess a new issue can be created if this is added to the config cat SDK.

test/OpenFeature.Contrib.ConfigCat.Test/ConfigCatProviderTest.cs
```diff
@@ -0,0 +1,101 @@
...
+
+        private static FlagOverrides BuildFlagOverrides(params (string key, object value)[] values)
+        {
+            var dictionary = new Dictionary<string, object>();
+            foreach (var (key, value) in values)
+            {
+                dictionary.Add(key, value);
+            }
+
+            return FlagOverrides.LocalDictionary(dictionary, OverrideBehaviour.LocalOnly);
```
toddbaert, line 98: This definitely makes testing this provider easy!

src/OpenFeature.Contrib.ConfigCat/ConfigCatProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+        /// <inheritdoc/>
+        public override async Task<ResolutionDetails<Value>> ResolveStructureValue(string flagKey, Value defaultValue, EvaluationContext context = null)
+        {
+            var stringDefaultValue = defaultValue?.AsString;
+            var result = await ProcessFlag(flagKey, context, stringDefaultValue);
+            var returnValue = result.Value == null ? defaultValue : new Value(result.Value);
+            return new ResolutionDetails<Value>(flagKey, returnValue, variant: result.Variant, errorMessage: result.ErrorMessage);
+        }
+
+        private async Task<ResolutionDetails<T>> ProcessFlag<T>(string flagKey, EvaluationContext context, T defaultValue)
```
toddbaert, line 69:
```suggestion
        private async Task<ResolutionDetails<T>> ResolveFlag<T>(string flagKey, EvaluationContext context, T defaultValue)

```

Nit: "resolve" is the term OpenFeature likes to use, so I'd recommend that name (though since this is a private function, it's really up to you.

src/OpenFeature.Contrib.ConfigCat/ConfigCatProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+        {
+            var user = context?.BuildUser();
+            var result = await Client.GetValueDetailsAsync(flagKey, defaultValue, user);
+            return new ResolutionDetails<T>(flagKey, result.Value, ParseErrorType(result.ErrorMessage), errorMessage: result.ErrorMessage, variant: result.VariationId);
```
toddbaert, line 73: With this SDK (and most others) the way to indicate an error is to throw an [exception](https://github.com/open-feature/dotnet-sdk-contrib/tree/main/src/OpenFeature.Contrib.Providers.GOFeatureFlag/exception). This is particularly important because [error hooks](https://openfeature.dev/specification/sections/hooks) only run if an error is thrown. I would amend this logic so that if an error is detected, you throw instead.

I understand this is a bit confusing. We should probably make it impossible to directly _return_ an error code in from the Resolvers.

src/OpenFeature.Contrib.ConfigCat/UserBuilder.cs
```diff
@@ -0,0 +1,51 @@
...
+using ConfigCat.Client;
+using OpenFeature.Model;
+
+namespace OpenFeature.Contrib.ConfigCat
+{
+    internal static class UserBuilder
+    {
+        private static readonly string[] PossibleUserIds = { "ID", "IDENTIFIER" };
+
+        internal static User BuildUser(this EvaluationContext context)
```
toddbaert, line 13: It seems this isn't covered by tests.

src/OpenFeature.Contrib.ConfigCat/UserBuilder.cs
```diff
@@ -0,0 +1,51 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using ConfigCat.Client;
+using OpenFeature.Model;
+
+namespace OpenFeature.Contrib.ConfigCat
+{
+    internal static class UserBuilder
+    {
+        private static readonly string[] PossibleUserIds = { "ID", "IDENTIFIER" };
```
toddbaert, line 11: It may be worth documenting how this mapping works in a README.

src/OpenFeature.Contrib.ConfigCat/UserBuilder.cs
```diff
@@ -0,0 +1,51 @@
...
+                    case "EMAIL":
+                        user.Email = value.Value.AsString;
+                        continue;
+                    case "COUNTRY":
+                        user.Country = value.Value.AsString;
+                        continue;
+                    default:
+                        user.Custom.Add(value.Key, value.Value.AsString);
+                        continue;
+                }
```
toddbaert, line 37: Same as https://github.com/open-feature/dotnet-sdk-contrib/pull/119/files#r1438292902

toddbaert (changes requested):
Wow @luizbon ! Thanks so much for this contribution, and sorry my review was delayed (holidays :sweat_smile: )

I left a few nits which are up to you. I think the only required changes from my perspective are to add a README and [this](https://github.com/open-feature/dotnet-sdk-contrib/pull/119/files#r1438290689).

I've also opened a PR to add you and @laliconfigcat to the owners file [here](https://github.com/prospa-group-oss/dotnet-sdk-contrib/pull/1).

src/OpenFeature.Contrib.ConfigCat/ConfigCatProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+        public override Metadata GetMetadata()
+        {
+            return new Metadata(Name);
+        }
+
+        /// <inheritdoc/>
+        public override async Task<ResolutionDetails<bool>> ResolveBooleanValue(string flagKey, bool defaultValue, EvaluationContext context = null)
+        {
+            return await ProcessFlag(flagKey, context, defaultValue);
+        }
```
adams85, line 40: No need for an async state machine here, we can just return the task returned by `ProcessFlag` directly as it's done in the case of the other `Resolve...` methods.

```suggestion
        public override Task<ResolutionDetails<bool>> ResolveBooleanValue(string flagKey, bool defaultValue, EvaluationContext context = null)
        {
            return ProcessFlag(flagKey, context, defaultValue);
        }

```

src/OpenFeature.Contrib.ConfigCat/ConfigCatProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+            return ProcessFlag(flagKey, context, defaultValue);
+        }
+
+        /// <inheritdoc/>
+        public override async Task<ResolutionDetails<Value>> ResolveStructureValue(string flagKey, Value defaultValue, EvaluationContext context = null)
+        {
+            var stringDefaultValue = defaultValue?.AsString;
+            var result = await ProcessFlag(flagKey, context, stringDefaultValue);
+            var returnValue = result.Value == null ? defaultValue : new Value(result.Value);
+            return new ResolutionDetails<Value>(flagKey, returnValue, variant: result.Variant, errorMessage: result.ErrorMessage);
```
adams85, line 66: This implementation looks problematic because this will work only when the feature flag value is a string (for more details, see https://configcat.com/docs/sdk-reference/dotnet/#setting-type-mapping).

However, `Value` can store values of other types than string (see [type definition](https://github.com/open-feature/dotnet-sdk/blob/e8ca1da9ed63df9685ec49a9569e0ec99ba0b3b9/src/OpenFeature/Model/Value.cs)).

To make this work with all the feature flag types which is supported by ConfigCat (boolean, string, int, double), we need something like this:

```suggestion
            var user = context?.BuildUser();
            var result = await Client.GetValueDetailsAsync(flagKey, defaultValue?.AsObject, user);
            var returnValue = result.IsDefaultValue ? defaultValue : new Value(result.Value);
            return new ResolutionDetails<Value>(flagKey, returnValue, ParseErrorType(result.ErrorMessage), errorMessage: result.ErrorMessage, variant: result.VariationId);
```

test/OpenFeature.Contrib.ConfigCat.Test/ConfigCatProviderTest.cs
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        [Theory]
+        [AutoData]
+        public async Task GetStructureValue_ForFeature_ReturnExpectedResult(string sdkKey)
+        {
+            const string jsonValue = "{ \"key\": \"value\" }";
+            var defaultValue = new Value(jsonValue);
+            var configCatProvider = new ConfigCatProvider(sdkKey,
+                options => { options.FlagOverrides = BuildFlagOverrides(("example-feature", defaultValue)); });
```
adams85, line 69: Flag overrides will only work with values which can be converted to supported feature flag values (boolean, string, int, double). The ConfigCat SDK doesn't know about the `Value` type, doesn't know how to convert it to a feature flag value, so this will lead to an error.

Using a value of a supported feature flag type solves the issue:

```suggestion
                options => { options.FlagOverrides = BuildFlagOverrides(("example-feature", defaultValue.AsString)); });
```

adams85 (changes requested):
Hi, I'm Adam from ConfigCat, I'm the primary maintainer of the ConfigCat .NET SDK currently. This is why @laliconfigcat asked me to take a look at this PR.

Here are my findings.

src/OpenFeature.Contrib.ConfigCat/ConfigCatProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+        /// <summary>
+        /// Creates new instance of <see cref="ConfigCatProvider"/>
+        /// </summary>
+        /// <param name="sdkKey">SDK Key to access the ConfigCat config.</param>
+        /// <param name="configBuilder">The action used to configure the client.</param>
+        /// <exception cref="ArgumentNullException"><paramref name="sdkKey"/> is <see langword="null"/>.</exception>
+        /// <exception cref="ArgumentException"><paramref name="sdkKey"/> is an empty string or in an invalid format.</exception>
+        public ConfigCatProvider(string sdkKey, Action<ConfigCatClientOptions> configBuilder = null)
+        {
+            Client = ConfigCatClient.Get(sdkKey, configBuilder);
```
adams85, line 27: One more remark. `ConfigCatClient` implements `IDisposable` and instances of such types should usually be disposed by the consumer as soon as they don't need them any longer.

I'm not sure though how big an issue is this in the case of OpenFeature, what are your guidelines regarding resource management. Probably it's not an actual issue as when we were implementing `ConfigCatClient`, we paid careful attention that client instances can be collected by the GC when no strong references to them are left.

test/OpenFeature.Contrib.ConfigCat.Test/ConfigCatProviderTest.cs
luizbon: Kudos to the ConfigCat team.

src/OpenFeature.Contrib.ConfigCat/ConfigCatProvider.cs
luizbon: Thanks for picking that up.

src/OpenFeature.Contrib.ConfigCat/ConfigCatProvider.cs
luizbon: For what it is worth, I have implemented the `IDisposable` interface in the provider class to handle this correctly.

src/OpenFeature.Contrib.ConfigCat/ConfigCatProvider.cs
luizbon: I left it as is for now since it'll capture the general exception anyway.

src/OpenFeature.Contrib.ConfigCat/UserBuilder.cs
luizbon: Added a README with that info.

src/OpenFeature.Contrib.ConfigCat/UserBuilder.cs
luizbon: Added tests for this class

src/OpenFeature.Contrib.ConfigCat/UserBuilder.cs
luizbon: Added a README with that info.

luizbon:
Hi @toddbaert and @adams85, sorry for the delay, I just came back from holidays.
I think I have addressed all your comments. Can you please review it again?

src/OpenFeature.Contrib.ConfigCat/ConfigCatProvider.cs
```diff
@@ -0,0 +1,118 @@
...
+            {
+                return ErrorType.TypeMismatch;
+            }
+            return ErrorType.General;
+        }
+
+        /// <summary>
+        /// Disposes the <see cref="Client"/>.
+        /// </summary>
+        public void Dispose()
```
adams85, line 113: After taking a closer look at the OpenFeature SDK architecture, specifically at [this PR](https://github.com/open-feature/dotnet-sdk/pull/158), it looks like making the provider class implement `IDisposable` is not the correct approach after all.

Also, a proper `Dispose` implementation would be tricky because of the way ConfigCat SDK manages client instances: 

> You can acquire singleton client instances for your SDK keys using the ConfigCatClient.Get(sdkKey: "<sdkKey>") static factory method. (However, please keep in mind that **subsequent calls to ConfigCatClient.Get() with the same SDK Key return a shared client instance, which was set up by the first call**.)

That is, if you create two provider instances with the same SDK key and then dispose one of them, the underlying client will be disposed in the other one as well. Probably we would need some kind of reference counting to make this work correctly.

So, in light of all this, I say we don't implement `IDisposable` for now. Let's address this issue when the provider shutdown feature lands.

luizbon: Yeah, I wasn't aware of the Shutdown implementation. I'll remove the disposable implementation.

adams85 (approved):

beeme1mr:
@toddbaert could you please review this tomorrow?

@luizbon could you please address the formatting issue when you have a moment?

luizbon:
Thanks for reminding me @beeme1mr, I have applied the format fix.

toddbaert (approved):
@luizbon Thanks again for your contribution, and your patience with the review process. I am approving, however, **I will request one more change** (I'd do it myself if I could, but I can't modify your branch): please add yourself (and perhaps @laliconfigcat and @adams85 if they are interested) to the [component_owners](https://github.com/open-feature/dotnet-sdk-contrib/blob/main/.github/component_owners.yml) for this component, and it's tests. This will add you to PRs for this component, and really helps share the maintenance load.

@laliconfigcat if you end up making changes to the config-cat SDK to send some kind of error code as you mention [here](https://github.com/open-feature/dotnet-sdk-contrib/pull/119#discussion_r1434556165), please create a PR or issue to improve the handling here if you dont mind! :pray:

luizbon:
Thanks, @toddbaert; just made the change to the component owner file and also found I needed to fix the path and package ID for the project.
It should be good to go now.

toddbaert:
@adams85 @laliconfigcat are either of you interested in being added to the `compoent_owners`? This will add your automatically to any PRs impacting this component. We only require one, so @luizbon 's addition is sufficient (note you will also have to be an org member for this automation to work).

In any case, I will merge this tomorrow EOD unless I hear objections! Thanks @luizbon !

laliconfigcat:
> @adams85 @laliconfigcat are either of you interested in being added to the `compoent_owners`? This will add your automatically to any PRs impacting this component. We only require one, so @luizbon 's addition is sufficient (note you will also have to be an org member for this automation to work).
> 
> In any case, I will merge this tomorrow EOD unless I hear objections! Thanks @luizbon !

Thanks for the suggestion, but I just jumped in for a quick PR review. I think @luizbon is the real component owner here :).

PR #120, opened 2023-12-21
fix: do not send targeting key as separate trait in flagsmith

vpetrusevici:
## This PR
- Removes targeting key (identity) from traits list to avoid unnecessary duplication


askpt (approved):
LGTM! This has a lot of casts but the `EvaluationContext` doesn't expose a method to remove a key.

toddbaert (approved):

PR #122, opened 2023-12-21
feat: Add custom dimensions for metrics

### LINKED ISSUE #118: [OTel Hook] Add custom-dimensions support to Metrics Hook
Support the addition of adding custom dimensions (attributes) to the `evaluation_success_total` metric. This is done by allowing users to define a function to extract dimensions from flag metadata.

[Here](https://github.com/open-feature/java-sdk-contrib/blob/main/hooks/open-telemetry/README.md#custom-dimensions-attributes-1) is the equivalent Java implementation.

askpt:
## This PR

- Adding custom dimensions for metrics

### Related Issues

Fixes #118 

### Notes

- Unfortunately, this feature has the same issue as the initial metrics. I cannot assert if the tags are added correctly. See #114 for more info.
- Removed the `otel` library from the unit tests so it will use the same one that is provided in the main project.

### How to test

On the yellow boxes you can see the custom metrics.
Screenshot 2023-12-22 at 13 16 40


src/OpenFeature.Contrib.Hooks.Otel/MetricsHook.cs
```diff
@@ -22,12 +22,16 @@ public class MetricsHook : Hook
...
 
         /// <summary>
         /// Initializes a new instance of the <see cref="MetricsHook"/> class.
         /// </summary>
-        public MetricsHook()
+        /// <param name="customDimensions">The optional custom dimensions.</param>
+        public MetricsHook(MetricHookCustomDimensions customDimensions = null)
         {
+            _customDimensionsTagList = customDimensions?.GetTagList() ?? Array.Empty<KeyValuePair<string, object>>();
+
```
toddbaert, line 34: The flexibility of this approach is a bit limited. I think a functional approach might be better. Instead of allowing users to define k/v pairs, I would recommend you allow users to define a function that takes an FlagEvaluationDetails object as a param, and returns a tuple or k/v set that will be added. Then users can build whatever data they want and add it as metics.

We do something similar in JS: https://github.com/open-feature/js-sdk-contrib/blob/main/libs/hooks/open-telemetry/README.md#custom-attributes

toddbaert:
@askpt I'm guessing based on your commits and the build status, you're still working on this, so I'll mark it draft for now. Feel free to mark it ready whenever.

src/OpenFeature.Contrib.Hooks.Otel/MetricsHook.cs
askpt: @toddbaert I was looking into the Java implementation and they offer a Flag Metadata (that is not implemented on dotnet SDK yet) and uses a custom dimensions that have a predefined list. Might be interesting to keep the extra dictionary and when we have the flag metadata implemented at SDK level, we use the FlagEvaluationDetails metadata?

toddbaert: Sorry I missed this comment for so long @askpt . I'm not sure I fully understand what you're proposing. Maybe you can clarify.

In any case, I think that once you implement the metadata [here](https://github.com/open-feature/dotnet-sdk/pull/223), the functional approach used in Java and JS should work.

PR #129, opened 2024-01-09
Adding a Provider implementation on top of the standard dotnet FeatureManagement system.

ericpattison:
This is currently dependent on a preview feature being added to the standard dotnet FeatureManagement library called [Variants](https://github.com/microsoft/FeatureManagement-Dotnet/tree/preview?tab=readme-ov-file#variants).

This new feature is required as with out it, features are either on or off. Variants allow feature flags to provide configuration based values.


beeme1mr:
@toddbaert @bacherfl could you please review this tomorrow?

src/OpenFeature.Contrib.Providers.FeatureManagement/FeatureManagementProvider.cs
```diff
@@ -0,0 +1,201 @@
...
+
+            TargetingContext targetingContext = new TargetingContext();
+
+            if (evaluationContext.ContainsKey("UserId"))
+            {
+                Value userId = evaluationContext.GetValue("UserId");
+                if (userId.IsString) targetingContext.UserId = userId.AsString;
+            }
+
+            if (evaluationContext.ContainsKey("Groups"))
```
bacherfl, line 130: nitpick: I suggest to add a constant for `UserId` and `Groups`

test/OpenFeature.Contrib.Providers.FeatureManagement.Test/FeatureManagementProviderTestNoContext.cs
```diff
@@ -0,0 +1,126 @@
...
+using Xunit;
+
+namespace OpenFeature.Contrib.Providers.FeatureManagement.Test
+{
+    public class FeatureManagementProviderTestNoContext
+    {
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public async Task MissingFlagKey_ShouldReturnDefault(bool defaultValue)
```
bacherfl, line 12: for the sake of completeness, I suggest to also add test cases for the other data types (i.e. `ResolveStringValue`, etc.) to check if the default value is returned in case the flag is not found in the configuration

src/OpenFeature.Contrib.Providers.FeatureManagement/OpenFeature.Contrib.Providers.FeatureManagement.csproj
```diff
@@ -0,0 +1,25 @@
+﻿﻿<Project Sdk="Microsoft.NET.Sdk">
```
bacherfl, line 1: There seems to have been a zero-width space character at the beginning of the file, which was preventing the build from succeeding. Not sure this will do the trick, but applying the suggestion below may fix that.

```suggestion
<Project Sdk="Microsoft.NET.Sdk">
```

ericpattison:
Updated to address @bacherfl's suggestions.

bacherfl (approved):
thanks for the additional tests! LGTM

bacherfl:
> thanks for the additional tests! LGTM

Seems like i did not see the failing dotnet-format check - can you please address that one? Other than that the implementation looks good to me

PR #132, opened 2024-01-11
refactor: Reduce OTel dependencies

austindrenski:
The `OpenFeature.Contrib.Hooks.Otel` project currently references (1) an exporter implementation which libraries should _not_ do under any normal circumstances, and (2) the main `OpenTelemetry` package which is excessive given all of the excellent work by the `dotnet/runtime` and `open-telemetry/opentelemetry-dotnet` teams have done over the last couple of years to ensure that the BCL types in `System.Diagnostics` conform to OTel standards.

This commit removes (1) and pares down (2) to from `OpenTelemetry` to `OpenTelemetry.Api` to shrink the dependency footprint we ship to consumers.

_At this point, I would argue that this hooks package should meet the minimal dependency requirement for upstream inclusion in the main OpenFeature package, but deferring that discussion for now._

### DIFF
--- a/src/OpenFeature.Contrib.Hooks.Otel/OpenFeature.Contrib.Hooks.Otel.csproj
+++ b/src/OpenFeature.Contrib.Hooks.Otel/OpenFeature.Contrib.Hooks.Otel.csproj
@@ -13,8 +13,7 @@
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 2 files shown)


beeme1mr (approved):

toddbaert (approved):

askpt (approved):
LGTM! These changes make complete sense 👍
Thank you

toddbaert (approved):

PR #136, opened 2024-01-16
chore: Enable Central Package Management (CPM)

austindrenski:
See: open-feature/dotnet-sdk#178

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -1,0 +1,39 @@
+<Project>
+
+  <PropertyGroup>
+    <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageVersion Include="ConfigCat.Client" Version="9.0.0" />
+    <PackageVersion Include="Flagsmith" Version="5.2.2" />
+    <PackageVersion Include="Google.Protobuf" Version="3.23.4" />
+    <PackageVersion Include="Grpc" Version="2.46.6" />
+    <PackageVersion Include="Grpc.Net.Client" Version="2.59.0" />
+    <PackageVersion Include="Grpc.Tools" Version="2.60.0" />
+    <PackageVersion Include="Microsoft.Extensions.Configuration.Json" Version="8.0.0" />
+    <PackageVersion Include="Microsoft.Extensions.DependencyInjection" Version="8.0.0" />
+    <PackageVersion Include="Microsoft.FeatureManagement" Version="4.0.0-preview" />
+    <PackageVersion Include="Microsoft.SourceLink.GitHub" Version="[1.0.0, 2.0)" />
+    <PackageVersion Include="OpenFeature" Version="[1.2,)" />
+    <PackageVersion Include="OpenTelemetry" Version="1.4.0" />
+    <PackageVersion Include="OpenTelemetry.Api" Version="1.4.0" />
+    <PackageVersion Include="OpenTelemetry.Exporter.InMemory" Version="1.4.0" />
+    <PackageVersion Include="System.Net.Http.WinHttpHandler" Version="8.0.0" />
+    <PackageVersion Include="System.Text.Json" Version="8.0.1" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageVersion Include="AutoFixture" Version="4.17.0" />
+    <PackageVersion Include="AutoFixture.Xunit2" Version="4.17.0" />
+    <PackageVersion Include="coverlet.collector" Version="3.1.2" />
+    <PackageVersion Include="coverlet.msbuild" Version="3.1.2" />
+    <PackageVersion Include="FluentAssertions" Version="6.7.0" />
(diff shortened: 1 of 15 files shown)


askpt (approved):
Always liked how clean the CPM makes the dependencies look like.

toddbaert (approved):

Directory.Packages.props
```diff
@@ -0,0 +1,39 @@
+<Project>
+
+  <PropertyGroup>
+    <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageVersion Include="ConfigCat.Client" Version="9.0.0" />
```
toddbaert, line 8: A potential downside of this is that "provider-unique" dependencies like this "ConfigCat.Client" are now centrally maintained. That's not ideal, particularly for such cases because it doesn't benefit from the ownerships defined [here](https://github.com/open-feature/dotnet-sdk-contrib/blob/main/.github/component_owners.yml).

Before this change, owners of the provider that uses `ConfigCat.Client` would be responsible for that dependency, and alerted to automated updates pertaining to it.

austindrenski: Hmmm, that's an interesting point!

I don't have a pre-canned solution, but I bet we could add a workflow that reacts to `dependabot`/`renovate` PRs and tags component owners if the dependency being updated is referenced by their component's `csproj`.

Assuming we can live with a minor coverage gap for a little while, I'll see if I can hack something together this weekend.

luizbon: For such cases we can omit the specific package from the `Directoty.Packages.props` and keep the reference only in the project file.
The only difference is the use of `VersionOverride` instead of `Version`.
So for ConfigCat provider, we would have this line in the csproj.
```
<PackageReference Include="ConfigCat.Client" VersionOverride="9.0.0" />
```

I don't know how bots will behave with such config though.

askpt: Very interesting and valid point. @austindrenski I remember playing around with the CPM and as far as I remember all projects within the solution needed to be using CPM as soon as it was enabled. Please correct me if I'm wrong.

Due to the nature of this repository, perhaps makes sense to keep the package versions managed by the project rather than a central place.

toddbaert: For dependencies like `ConfigCat.Client`, is there any reason we can't leave them only in the package that uses them, while maintaining only the common, less-specific deps centrally? This is basically what we do with the pom.xml setup in the java contrib.

luizbon: If we leave only the specific packages in each project, this PR adds no value, as we already have `Common.props` controlling versions centrally.

Is it worth moving to CPM?
I don't see any specific project referencing the same package outside `Common.props`.

We would have value if more projects started to use common packages, but not all of them. Even though those packages are not supposed to be used together, having different dependency versions could be fine.

austindrenski: @luizbon [wrote](https://github.com/open-feature/dotnet-sdk-contrib/pull/136#discussion_r1456416030) ___(emphasis added)___:

> [...]
>
> If we leave only the specific packages in each project, this PR adds no value, as ___we already have `Common.props` controlling versions centrally.___
> 
> Is it worth moving to CPM? I don't see any specific project referencing the same package outside `Common.props`.
>
> [...]

_Didn't see this before posting my previous reply, but wanted to respond to this specifically:_

I hear where you're coming from on this, but when it comes to repo engineering, I think there's value in using standard mechanisms whenever possible, even if they don't necessarily provide "more" than what our custom mechanism does.

So a couple of years ago, I would've been all for the `Common.props` pattern, because it definitely beats tracking versions in each `csproj`, but in a post-CPM world, I have a growing preference for just using the built-in features where possible/practical.

luizbon: This is a good discussion @austindrenski.

Either approach will achieve the goal.

However, I don't see the use of `Common.props` as a non-standard way of tracking package versions. It is a supported solution to stop the repetition across projects, and the version is one of them.

Indeed, CPM is a more elegant/modern solution for package control, but this specific project structure adds more complexity than solves the problem.
Adding more customisation to the workflow is proof of that.

Sorry to say, but I still don't see value in adding CPM to this repo.

vpetrusevici (approved):

Directory.Packages.props
austindrenski: Closing the loop here: yes, you are all correct, and I'm afraid I just outright misunderstood the governance of this repo 🤦🤦🤦

Given that this repo is for co-hosting/-locating contributor-maintained projects that should be able to evolve and release independently, this PR is indeed the wrong approach.

Thanks for your patience while I caught up with what you were saying @luizbon and @toddbaert 🙇🙇🙇

austindrenski:
Closing after discussion because dependency isolation is by-design in contrib-style repos _(and possibly something we should take further w.r.t. moving all dependency versions out of `Common.prod.props`)_.

See: https://github.com/open-feature/dotnet-sdk-contrib/pull/136#pullrequestreview-1827964995, https://github.com/open-feature/dotnet-sdk-contrib/pull/136#discussion_r1465316395

PR #144, opened 2024-01-23
chore: Bump OpenFeature to 1.4.0 + react to deprecations

austindrenski:
See: open-feature/dotnet-sdk#154, open-feature/dotnet-sdk#170, open-feature/dotnet-sdk#171

### DIFF
--- a/build/Common.props
+++ b/build/Common.props
@@ -25,7 +25,7 @@
     -->
     <MicrosoftSourceLinkGitHubPkgVer>[1.0.0,2.0)</MicrosoftSourceLinkGitHubPkgVer>
     <!-- 0.5+ -->
-    <OpenFeatureVer>[1.2,)</OpenFeatureVer>
+    <OpenFeatureVer>[1.4,)</OpenFeatureVer>
   </PropertyGroup>
 
   <ItemGroup Condition="'$(OS)' == 'Unix'">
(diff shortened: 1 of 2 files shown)


build/Common.props
```diff
@@ -25,7 +25,7 @@
     -->
     <MicrosoftSourceLinkGitHubPkgVer>[1.0.0,2.0)</MicrosoftSourceLinkGitHubPkgVer>
     <!-- 0.5+ -->
-    <OpenFeatureVer>[1.2,)</OpenFeatureVer>
+    <OpenFeatureVer>1.4.0</OpenFeatureVer>
```
askpt, line 28:
```suggestion
    <OpenFeatureVer>[1.4,)</OpenFeatureVer>
```

Should we keep the 1.4 version as minor and let it bump naturally? Also, we might need to bump again when 2.0 comes with the breaking changes.

askpt (commented):
Small question. LGTM otherwise.

build/Common.props
toddbaert: This makes sense to me

toddbaert (approved):

build/Common.props
austindrenski: IMO floating package versions haunt my nightmares. I've been burned more than once by a floating package version causing a lower version to get resolved from a CI build server cache.

That doesn't really apply in the era of cloud-hosted CI servers though, so will revert...

askpt (approved):
🚀

Issue #139, opened 2024-01-17
[flagd] implement in-process mode


toddbaert:
As mentioned in flagd documentation of  **Creating an in-process flagd provider** [^1], the in-process provider comes with unique advantages. This issue focuses on creating a .NET implementation.
**Requirements**

- Fulfill requirements outlined by flagd in-process guideline [^1]
- Minimal external dependencies 
- Updated/New README & usage guide
- Unit & e2e tests (e2e test definitions exist [here](https://github.com/open-feature/flagd-testbed/tree/main/gherkin))

**Technical requirements**

- Build both remote and in-process providers into a single dependency (To be decided after the POC)
- Reuse existing interfaces and implementations where possible 

**Tasks**

- [x] JSON logic implementation (including custom ops)
- [x] Provider functionality (sync.proto implementation)
- [x] e2e tests for both RPC and in-process
  - e2e test already exist [here](https://github.com/open-feature/dotnet-sdk/tree/main/test/OpenFeature.E2ETests) for the SDK, start with these

[^1]: https://flagd.dev/reference/specifications/in-process-providers/

toddbaert:
Done with https://github.com/open-feature/dotnet-sdk-contrib/pull/189

PR #149, opened 2024-01-29
feat: in-process provider

bacherfl:
Part of #139

This PR adds the support for an in-process provider, using the [flagd.sync](https://github.com/open-feature/flagd-schemas/blob/json/json-schema-v0.2.0/protobuf/flagd/sync/v1/sync.proto) API to connect to a sync stream serving the flag configurations specified by the selector (determined by the `FLAGD_SOURCE_SELECTOR` environment variable).

Currently only the built-in JSONLogic operators are supported, i.e. the custom operators specified [here](https://flagd.dev/reference/specifications/in-process-providers/#custom-jsonlogic-evaluators) are yet to be implemented - however, it might be better to implement this in a follow-up PR


bacherfl:
This should be ready fir a first round of reviews - @toddbaert as discussed, we can leave this PR open and create a separate PR on top of that to implement the custom JSONLogic evaluators, or merge this one without the custom evaluators and deactivate the in-process mode until the additional evaluators have been implemented - whatever you prefer :)

toddbaert:
Thanks @bacherfl ! I will thoroughly review this in the next couple days!

src/OpenFeature.Contrib.Providers.Flagd/FlagdProvider.cs
```diff
@@ -28,16 +28,11 @@ namespace OpenFeature.Contrib.Providers.Flagd
...
         const string ProviderName = "flagd Provider";
-        static int EventStreamRetryBaseBackoff = 1;
         private readonly FlagdConfig _config;
-        private readonly Service.ServiceClient _client;
-        private readonly Metadata _providerMetadata = new Metadata(ProviderName);
 
-        private readonly ICache<string, object> _cache;
-        private int _eventStreamRetries;
-        private int _eventStreamRetryBackoff = EventStreamRetryBaseBackoff;
+        private readonly Metadata _providerMetadata = new Metadata(ProviderName);
```
toddbaert, line 33: It's a really good idea to refactor a lot of code out of this file, nice job. There's a lot of unused imports at the top now though.

src/OpenFeature.Contrib.Providers.Flagd/FlagdConfig.cs
```diff
@@ -50,6 +71,18 @@ internal int MaxEventStreamRetries
...
+            get => _resolverType;
+            set => _resolverType = value;
+        }
+
+        internal string SourceSelector
+        {
+            get { return _sourceSelector; }
+            set { _sourceSelector = value; }
+        }
+
```
toddbaert, line 85: In Kubernmetes, the ENV_VAR based configs make a lot of sense, but I think we need a way to use the in-process provider in code as well, so users who aren't un K8s can use whatever configuration tools they want to configure flagd directly through code.

I think we should either allow configuration through `FlagdConfig` (which would mean exposing a public constructor for it, with public fields) or add another constructor on the provider that takes a URL and Resolver type. Whatever is easier, I think (though I should point out we also have a public `FlagdProvider` constructor which takes a `FlagdConfig` which is sort-of useless now I think.

Kavindu-Dodan: +1 for exposing a configuration-based constructor. This should be easily done through a builder. And I can point to Java [1] & go flagd [2] providers which allows both env vars and constructor paramters 

[1] - https://github.com/open-feature/java-sdk-contrib/tree/main/providers/flagd#in-process-resolver 
[2] - https://github.com/open-feature/go-sdk-contrib/tree/main/providers/flagd#in-process-resolver

src/OpenFeature.Contrib.Providers.Flagd/InProcessResolver.cs
```diff
@@ -0,0 +1,276 @@
...
+using System.Net.Security;
+using System.Security.Cryptography.X509Certificates;
+using Grpc.Net.Client;
+using OpenFeature.Flagd.Grpc;
+using System.Net.Sockets;
+using System.Threading;
+using Grpc.Core;
+using Value = OpenFeature.Model.Value;
+
+namespace OpenFeature.Contrib.Providers.Flagd
```
Kavindu-Dodan, line 17: minor - may be it's good to namespace in-process vs rpc so that they are proper access modifiers. In the long run this should help in organizing, especially with upcoming json evaluators, offline mode file readers sort of logic.

src/OpenFeature.Contrib.Providers.Flagd/InProcessResolver.cs
```diff
@@ -0,0 +1,276 @@
...
+        }
+
+        public Task<ResolutionDetails<Value>> ResolveStructureValue(string flagKey, Value defaultValue, EvaluationContext context = null)
+        {
+            return Task.FromResult(_evaluator.ResolveStructureValue(flagKey, defaultValue, context));
+        }
+
+        private async void HandleEvents()
+        {
+            while (_eventStreamRetries < _config.MaxEventStreamRetries)
```
Kavindu-Dodan, line 86: I think this needs to be improved. In-process provider is special as it only have a single flag sync mechanism (currently grpc stream). If we end reconnections after few retries, then flags store will go stale and this is not a desired state. Similar to what we do in other implementations, (for example Java [1]) we should use some exponential back-off (up to a limit) and continuously re-attempt to connect till we get a shutdown request. This should help with debugging as there will be continuous errors if something is wrong with connectivity (ex:- api address change) and reconnect and server fresh flags if problem goes away 

[1] - https://github.com/open-feature/java-sdk-contrib/blob/main/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/FlagStore.java#L88

src/OpenFeature.Contrib.Providers.Flagd/JsonEvaluator.cs
```diff
@@ -0,0 +1,223 @@
...
+            return ResolveValue(flagKey, defaultValue, context);
+        }
+
+        private ResolutionDetails<T> ResolveValue<T>(string flagKey, T defaultValue, EvaluationContext context = null)
+        {
+            // check if we find the flag key
+            var reason = Reason.Default;
+            if (_flags.TryGetValue(flagKey, out var flagConfiguration))
+            {
+                reason = Reason.Static;
```
Kavindu-Dodan, line 124: We have to check for `DISABLED` state of the flag. Maybe you can refer Java impl [1] or go impl [2] to see if all checks map here :) 

[1] - https://github.com/open-feature/java-sdk-contrib/blob/main/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java#L158

[2] - https://github.com/open-feature/flagd/blob/main/core/pkg/evaluator/json.go#L292

src/OpenFeature.Contrib.Providers.Flagd/InProcessResolver.cs
toddbaert: Also worth noting that some gRPC implementations have listener implementations so you can listen to the underlying connection state and restore the stream when the connection is restored. With this model, the stream retry options etc are used to control the parameters of the underlying connection retries, and the stream just reconnects accordingly.

toddbaert: It looks like channel.WaitForStateChangedAsync could perhaps help us here, but its not available in all our targets :weary:

Kavindu-Dodan: @toddbaert  yeah, that's a good suggestion. And I hope unlike Java, .Net exposes configurations to control internal grpc reconnect configurations

toddbaert:
With the [new testbed release](https://github.com/open-feature/flagd-testbed/releases) and some hacks, I was able to get the test suite working for 9/11 test cases. I've left a few comments.

src/OpenFeature.Contrib.Providers.Flagd/InProcessResolver.cs
toddbaert, line 86: By default `while (_eventStreamRetries < _config.MaxEventStreamRetries)` evaluates to `False`, since `MaxEventStreamRetries` is zero (this means nothing is ever read). Additionally the retries ENV VAR seem to be ignored without the cache.

src/OpenFeature.Contrib.Providers.Flagd/InProcessResolver.cs
```diff
@@ -0,0 +1,276 @@
...
+            _mtx = new Mutex();
+            _evaluator = new JsonEvaluator(config.SourceSelector);
+        }
+
+
+        public void Init()
+        {
+            var handleEvents = new Thread(HandleEvents);
+            handleEvents.Start();
+        }
```
toddbaert, line 52: This method needs to block until the connection is established, or else the setProvider returns and the READY event fires before the provider is actually READY (and indeed in might never bet ready). I would use some kind of latch or callback or something to block until we connect.

As a hacky fix you can add a Thread.sleep here for 1000 and things "work".

src/OpenFeature.Contrib.Providers.Flagd/InProcessResolver.cs
```diff
@@ -0,0 +1,276 @@
...
+
+        public void Shutdown()
+        {
+            throw new NotImplementedException();
```
toddbaert, line 56: We need to stop the threads and kill the connections here. As it is, with every SetProvider for this provider we have resource leaks.

We need to do the same thing in the RPC version.

toddbaert (changes requested):
With the [new testbed release](https://github.com/open-feature/flagd-testbed/releases) and some hacks, I was able to get the test suite working for 9/11 standard test cases (these are standard test from the OpenFeature spec itself and don't include custom flagd evaluators). I've left a few comments.

The 2 test cases that fail are [these](https://github.com/open-feature/spec/blob/738bb41a3e5002d037d4ef91f71a7d0c7d3893c3/specification/assets/gherkin/evaluation.feature#L58-L67). the first fails because it asserts the reason is `ERROR` and it's `DEFAULT` and the second because it expects a `TYPE_MISMATCH` error code and gets a `FLAG_NOT_FOUND`. I guess both of these are two do minor logic flaws in the Resolve method.

src/OpenFeature.Contrib.Providers.Flagd/JsonEvaluator.cs
```diff
@@ -0,0 +1,223 @@
...
+                        );
+                }
+                reason = Reason.Default;
+            }
+            return new ResolutionDetails<T>(
+                flagKey: flagKey,
+                value: defaultValue,
+                reason: reason,
+                errorType: ErrorType.FlagNotFound
+                );
```
toddbaert, line 169: Don't return error codes, throw `FlagNotFound` instead (this causes the error handlers to run).

I know this is confusing. I'd like to release a new SDK version at some point that prevents providers from returning errors like this (forcing throws).

src/OpenFeature.Contrib.Providers.Flagd/JsonEvaluator.cs
```diff
@@ -0,0 +1,223 @@
...
+                            foundVariantValue = Convert.ToInt32(foundVariantValue);
+                        }
+                    }
+                    else if (foundVariantValue is JObject value)
+                    {
+                        foundVariantValue = ConvertJObjectToOpenFeatureValue(value);
+                    }
+                    return new ResolutionDetails<T>(
+                        flagKey: flagKey,
+                        value: (T)foundVariantValue,
```
toddbaert, line 157: I think you need to add a check here before this cast for type mismatches (and throw a `FlagTypeError` if it's irrational).

src/OpenFeature.Contrib.Providers.Flagd/README.md
```diff
@@ -107,3 +109,48 @@ var flagdProvider = new FlagdProvider(new Uri("http://localhost:8013"));
...
+
+            // set the host and port for flagd-proxy
+            Environment.SetEnvironmentVariable("FLAGD_HOST", "localhost");
+            Environment.SetEnvironmentVariable("FLAGD_PORT", "8015");
+            // set the resolver type to 'IN_PROCESS'
+            Environment.SetEnvironmentVariable("FLAGD_RESOLVER_TYPE", "IN_PROCESS");
+            // provide the flag source selector, e.g. the name of a Flags custom resource which is watched by the flagd-proxy
+            Environment.SetEnvironmentVariable("FLAGD_SOURCE_SELECTOR", "core.openfeature.dev/flags/sample-flags");
+
+            var flagdProvider = new FlagdProvider();
```
Kavindu-Dodan, line 141: Minor nit - example could be updated to use builder and config bound constructor. This then matches with the rest of provider documentations :)

src/OpenFeature.Contrib.Providers.Flagd/Resolver/Rpc/RpcResolver.cs
```diff
@@ -0,0 +1,494 @@
...
+            _client = BuildClientForPlatform(_config);
+            _mtx = new System.Threading.Mutex();
+
+            if (_config.CacheEnabled)
+            {
+                _cache = new LRUCache<string, object>(_config.MaxCacheSize);
+            }
+        }
+
+        internal RpcResolver(Service.ServiceClient client, FlagdConfig config, ICache<string, object> cache = null)
```
Kavindu-Dodan, line 49: Minor - from quick look on the usage, this seems to be not used. If it does however, I think it's best to constructor chain to this constructor from config based constructor :)

src/OpenFeature.Contrib.Providers.Flagd/Resolver/InProcess/InProcessResolver.cs
```diff
@@ -0,0 +1,288 @@
...
+
+        internal InProcessResolver(FlagdConfig config)
+        {
+            _config = config;
+            _client = BuildClient(config, channel => new FlagSyncService.FlagSyncServiceClient(channel));
+            _mtx = new Mutex();
+            _evaluator = new JsonEvaluator(config.SourceSelector);
+        }
+
+        internal InProcessResolver(FlagSyncService.FlagSyncServiceClient client, FlagdConfig config)
```
Kavindu-Dodan, line 42: Minor - please check if this is used (seems not per my view). And if it is, let's use constructor chaining to avoid duplicate code :)

src/OpenFeature.Contrib.Providers.Flagd/JsonEvaluator.cs
```diff
@@ -0,0 +1,226 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+using JsonLogic.Net;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using OpenFeature.Constant;
+using OpenFeature.Error;
+using OpenFeature.Model;
+
+namespace OpenFeature.Contrib.Providers.Flagd
```
Kavindu-Dodan, line 13: Minor - packing could be set to `Resolver.InProcess`

src/OpenFeature.Contrib.Providers.Flagd/Resolver/InProcess/InProcessResolver.cs
```diff
@@ -0,0 +1,288 @@
...
+using System.Linq;
+using System.Net.Security;
+using System.Security.Cryptography.X509Certificates;
+using Grpc.Net.Client;
+using System.Net.Sockets;
+using System.Threading;
+using Grpc.Core;
+using Value = OpenFeature.Model.Value;
+
+namespace OpenFeature.Contrib.Providers.Flagd.RResolver.InProcess
```
Kavindu-Dodan, line 16: Minor - `RResolver` could be a type

src/OpenFeature.Contrib.Providers.Flagd/Resolver/Rpc/RpcResolver.cs
bacherfl: good catch - it is used in the unit tests, but i adapted the code to use a constructor chain :)

src/OpenFeature.Contrib.Providers.Flagd/Resolver/InProcess/InProcessResolver.cs
bacherfl: good catch - it is used in the unit tests, but i adapted the code to use a constructor chain :)

src/OpenFeature.Contrib.Providers.Flagd/Resolver/InProcess/InProcessResolver.cs
```diff
@@ -97,14 +97,13 @@ private async void HandleEvents(CountdownEvent latch)
                     // Read the response stream asynchronously
                     while (await call.ResponseStream.MoveNext(token))
                     {
+                        var response = call.ResponseStream.Current;
+                        _evaluator.Sync(FlagConfigurationUpdateType.ALL, response.FlagConfiguration);
```
toddbaert, line 101: I think this was a race - we we signaling the latch before we read, which caused some issues.

src/OpenFeature.Contrib.Providers.Flagd/FlagdProvider.cs
```diff
@@ -131,9 +107,24 @@ public static string GetProviderName()
...
+            {
+                this._resolver.Init();
+                this._status = ProviderStatus.Ready;
+
+            }).ContinueWith((t) =>
+            {
+                this._status = ProviderStatus.Error;
+                if (t.IsFaulted) throw t.Exception;
+            });
+        }
```
toddbaert, line 127: @bacherfl @Kavindu-Dodan I added an initialize function, the old provider was built before this was implemented; it's no longer done in the contstructor.

toddbaert (approved):
e2e tests are working with some minor noted changes. Thanks @bacherfl

src/OpenFeature.Contrib.Providers.Flagd/Resolver/Rpc/RpcResolver.cs
```diff
@@ -0,0 +1,492 @@
...
+            if (_config.CacheEnabled)
+            {
+                var handleEvents = new Thread(HandleEvents);
+                handleEvents.Start();
+            }
+        }
+
+        public void Shutdown()
+        {
+            throw new System.NotImplementedException();
```
Kavindu-Dodan, line 66: I think we should handle this as we now support init and shutdown

src/OpenFeature.Contrib.Providers.Flagd/Resolver/Rpc/RpcResolver.cs
```diff
@@ -0,0 +1,492 @@
...
+            _cache = cache;
+        }
+
+        public void Init()
+        {
+            if (_config.CacheEnabled)
+            {
+                var handleEvents = new Thread(HandleEvents);
+                handleEvents.Start();
+            }
```
Kavindu-Dodan, line 61: I think this is something interesting to consider for other implementations 🤔

Kavindu-Dodan (approved):
Left few improvement comments, but this already looks promising :)

src/OpenFeature.Contrib.Providers.Flagd/Resolver/Rpc/RpcResolver.cs
toddbaert: I've implemented this now, see commit [here](https://github.com/open-feature/dotnet-sdk-contrib/pull/149/commits/a0e86661f7dc3744e12a3a05f0a37f02a3d16b97).

cc @bacherfl

Kavindu-Dodan (approved):

PR #163, opened 2024-03-05
feat: Statsing provider

jenshenneberg:
## This PR
- Adding a minimal Statsig provider integration

### Notes
The implementation is minimal - only `ResolveBooleanValue` is implemented.

### How to test
Create a free Statsig account and use as described in README.md


src/OpenFeature.Contrib.Providers.Statsig/EvaluationContextExtensions.cs
```diff
@@ -0,0 +1,70 @@
...
+
+        public static StatsigUser AsStatsigUser(this EvaluationContext evaluationContext)
+        {
+            if (evaluationContext == null)
+                return null;
+
+            var user = new StatsigUser();
+            foreach (var item in evaluationContext)
+            {
+                //TODO: Await release containing this https://github.com/open-feature/dotnet-sdk/pull/231 to use TargetingKey instead of UserId
```
beeme1mr, line 26: I'm checking on this to see if we can release it soon.

src/OpenFeature.Contrib.Providers.Statsig/StatsigProvider.cs
```diff
@@ -0,0 +1,107 @@
...
+
+        /// <inheritdoc/>
+        public override Metadata GetMetadata() => _providerMetadata;
+
+        /// <inheritdoc/>
+        public override Task<ResolutionDetails<bool>> ResolveBooleanValue(string flagKey, bool defaultValue, EvaluationContext context = null)
+        {
+            //TODO: defaultvalue = true not yet supported due to https://github.com/statsig-io/dotnet-sdk/issues/33
+            if (defaultValue == true)
+                throw new FeatureProviderException(ErrorType.General, "defaultvalue = true not supported (https://github.com/statsig-io/dotnet-sdk/issues/33)");
```
beeme1mr, line 54: Will the Statsig SDK throw if you pass a default value of `true`? The OpenFeature SDK will catch any exception thrown and return the default value. It may not be necessary to explicitly check and throw here.

beeme1mr (approved):
Overall, it looks good. I left a few comments but I think we can merge this when you're ready.

src/OpenFeature.Contrib.Providers.Statsig/EvaluationContextExtensions.cs
jenshenneberg: Thank you, if it's close we might as well wait for that.

src/OpenFeature.Contrib.Providers.Statsig/StatsigProvider.cs
jenshenneberg: The Statsig SDK does not have a fallback option for boolean evaluation, I've created a ticket for it [here](https://github.com/statsig-io/dotnet-sdk/issues/33). 
I felt it would be dangerous to allow passing in a default value of `true` as the SDK will always return false for non defined feature gates.

jenshenneberg:
Thanks for having a look. 😄   I'll do some more testing and wait for a decision on a new release of https://github.com/open-feature/dotnet-sdk/ before merging

src/OpenFeature.Contrib.Providers.Statsig/EvaluationContextExtensions.cs
beeme1mr: We released 1.5 last night.

jenshenneberg: Thanks, @beeme1mr. Would you be ok with me doing the update [here](https://github.com/open-feature/dotnet-sdk-contrib/blob/533dfa652011b645d48c5d17dbdfe39b64a8eadf/build/Common.props#L27) as part of this PR, or does it need to be in a separate one?

beeme1mr: Sure, that would be fine with me.

jenshenneberg: In the end I've decided to include it in this PR as it was a minor update. Please let me know if you would like me to change that so it goes in its own PR.

jenshenneberg:
Updated to `OpenFeature` 1.5.* as part of this PR. Please merge this PR if you feel it's ready. Thanks for your input and help with this. 😃

PR #177, opened 2024-04-10
fix: Use new Statsig Api to return default value when flag is not defined

jenshenneberg:
## This PR
- updates the Statsig provider so it now returns the default value when a flag is not defined

### Notes
Fix was made possible after the Statsig added the new `GetFeatureGate` API [here](https://github.com/statsig-io/dotnet-sdk/commit/f3ddca04360555fa98974c8837fed78624d557f1)


askpt (approved):
LGTM!

beeme1mr (approved):

PR #178, opened 2024-04-10
feat: add Flipt provider

dmitryrogov:
## This PR
Adding Flipt provider integration.

### How to test
Run Flipt feature server locally
```bash
docker run -d  -p 8080:8080  -p 9000:9000 -v $HOME/flipt:/var/opt/flipt docker.flipt.io/flipt/flipt:latest
```


beeme1mr:
Hey @dmitryrogov, thanks for the PR. @toddbaert, @askpt, or @markphelps could someone please review this PR?

src/OpenFeature.Contrib.Providers.Flipt/README.md
```diff
@@ -0,0 +1,195 @@
...
+
+var builder = WebApplication.CreateBuilder(args);
+
+// Get provider config from configuration
+var config = builder.Configuration.GetSection("OpenFeature:Provider:Flipt").Get<FliptProviderConfiguration>();
+
+/* Or create config directly:
+var config = new FliptProviderConfiguration
+{
+    ServiceUri = new Uri("http://localhost:9000"),
```
markphelps, line 68: I think we'll want to support authentication as well, likely with some helpers that can add either clientToken authentication (https://docs.flipt.io/authentication/using-tokens) or JWT authentication (https://docs.flipt.io/authentication/using-jwts)

The only difference really is the value of the `Authorization` header that is sent to Flipt:
- Client Token: `Authentication: Bearer {token}`
- JWT: `Authentication: JWT {token}`

src/OpenFeature.Contrib.Providers.Flipt/FliptProvider.cs
```diff
@@ -0,0 +1,155 @@
...
+
+        /// <summary>
+        /// Creates new instance of <see cref="FliptProvider" />
+        /// </summary>
+        /// <param name="configuration">Flipt provider options.</param>
+        public FliptProvider(FliptProviderConfiguration configuration)
+        {
+            _configuration = configuration;
+            var channel = GrpcChannel.ForAddress(configuration.ServiceUri);
+            _client = new EvaluationServiceClient(channel);
```
markphelps, line 31: we may want to think about making a HTTP client and making that the default as GRPC requires HTTP2 and might not work in all environments.

I realize making an HTTP C# client for Flipt is not a small undertaking though, we've had this issue open for a bit in our server-side SDK monorepo: https://github.com/flipt-io/flipt-server-sdks/issues/86

markphelps (commented):
thank you for taking this on @dmitryrogov !!

src/OpenFeature.Contrib.Providers.Flipt/FliptProvider.cs
dmitryrogov: That's true. It would be great to use a client interface instead of the implementation itself (GRPC client SDK, HTTP client SDK or server SDK), but the library doesn't support that at the moment.
For now I can I suggest either modifying the repository [flipt-grpc-dotnet](https://github.com/flipt-io/flipt-grpc-dotnet) by adding a GRPC client interface, or making [this constructor](https://github.com/open-feature/dotnet-sdk-contrib/blob/81607213b7084e27682273cb3c8d5032036812dc/src/OpenFeature.Contrib.Providers.Flipt/FliptProvider.cs#L39) public

markphelps: Yeah we actually do something similar in our Go Open Feature provider, which supports both GRPC and HTTP

There its all based on the protocol in the address / URL string for which transport to use: https://github.com/open-feature/go-sdk-contrib/tree/main/providers/flipt#configuration

https://github.com/open-feature/go-sdk-contrib/blob/main/providers/flipt/pkg/service/transport/service.go#L152

We could do something similar here, where at first we would only support the GRPC transport like you have done, but once we ideally have a C# REST/HTTP SDK then we can add support here in a non-breaking way, by checking the protocol.

toddbaert:
Sorry I've been slow on this one. I'll get to it first thing Monday.

src/OpenFeature.Contrib.Providers.Flipt/version.txt
```diff
@@ -0,0 +1 @@
+0.0.1
```
toddbaert, line 1: Also add an entry in the [release-plz-manifest](https://github.com/open-feature/dotnet-sdk-contrib/blob/main/.release-please-manifest.json) with a version number matching this, or it will release the first version as 1.0.0.

toddbaert (approved):
Sorry this took so long, this looks good to me, and I really tried to find something wrong. If @markphelps has no objections I'm OK with merging. We can also add it to the [ecosystem](https://openfeature.dev/ecosystem/).

One important note about [this](https://github.com/open-feature/dotnet-sdk-contrib/pull/178/files#r1603581826), though.

src/OpenFeature.Contrib.Providers.Flipt/README.md
markphelps: also support for our authentication methods still needs to be added @dmitryrogov

src/OpenFeature.Contrib.Providers.Flipt/FliptProvider.cs
markphelps: I think this still needs addressing @dmitryrogov

src/OpenFeature.Contrib.Providers.Flipt/README.md
```diff
@@ -0,0 +1,195 @@
...
+var client = OpenFeature.Api.Instance.GetClient("my-app");
+
+// Resolve boolean flag value
+var isTestFlagEnabled = await client.GetBooleanValue("test-flag", false, null);
+
+System.Console.WriteLine($"Test flag enabled: {isTestFlagEnabled}");
+```
+
+## Boolean flag evaluation
+To perform Boolean flag evaluation, such as `GetBooleanValue` or `GetBooleanDetail` you can utilize [Boolean Evaluation](https://docs.flipt.io/reference/evaluation/boolean-evaluation) or [Variant Evaluation](https://docs.flipt.io/reference/evaluation/variant-evaluation) APIs.
```
markphelps, line 91: this seems a bit odd, wouldnt we want to rely on boolean flags always for `GetBooleanX` calls ?

markphelps (changes requested):
i think theres still a few items that could use some 👀 before merging

beeme1mr:
Closing in favor of https://github.com/open-feature/dotnet-sdk-contrib/pull/178

PR #186, opened 2024-04-23
fix(flagsmith/tests): Fix ValueError in FlagsmithProvider tests

matthewelwell:
## This PR

Fixes the unit tests against the Flagsmith provider which set up the client with an empty environment key. This feels like a breaking change, however, the client would not work with an empty environment key so shouldn't cause any harm. 

Happy to discuss further if needed. 

### Related Issues

https://github.com/open-feature/dotnet-sdk-contrib/pull/184

### How to test

```
dotnet test
```


askpt (approved):
LGTM!

PR #201, opened 2024-05-13
refactor: Use the newly introduced ConfigCat error codes to determine error type

adams85:
## This PR

ConfigCat .NET SDK has recently [introduced error codes](https://github.com/configcat/.net-sdk/releases/tag/v9.2.0) to allow consumers to detect the reason for errors that occur during evaluation in a more reliable way than guessing it by matching error message strings.

This PR upgrades the OpenFeature provider for ConfigCat to the new SDK version and adjusts the provider implementation to leverage this new feature.

### DIFF
--- a/src/OpenFeature.Contrib.Providers.ConfigCat/OpenFeature.Contrib.Providers.ConfigCat.csproj
+++ b/src/OpenFeature.Contrib.Providers.ConfigCat/OpenFeature.Contrib.Providers.ConfigCat.csproj
@@ -16,6 +16,6 @@
     </AssemblyAttribute>
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="ConfigCat.Client" Version="[9,)"/>
+    <PackageReference Include="ConfigCat.Client" Version="9.2.0"/>
   </ItemGroup>
 </Project>
--- a/src/OpenFeature.Contrib.Providers.ConfigCat/ConfigCatProvider.cs
+++ b/src/OpenFeature.Contrib.Providers.ConfigCat/ConfigCatProvider.cs
@@ -64,7 +64,7 @@ namespace OpenFeature.Contrib.ConfigCat
             var user = context?.BuildUser();
             var result = await Client.GetValueDetailsAsync(flagKey, defaultValue?.AsObject, user);
             var returnValue = result.IsDefaultValue ? defaultValue : new Value(result.Value);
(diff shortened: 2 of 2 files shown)


askpt (approved):

src/OpenFeature.Contrib.Providers.ConfigCat/OpenFeature.Contrib.Providers.ConfigCat.csproj
```diff
@@ -16,6 +16,6 @@
     </AssemblyAttribute>
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="ConfigCat.Client" Version="[9,)"/>
+    <PackageReference Include="ConfigCat.Client" Version="9.2.0"/>
```
beeme1mr, line 19: Could a range be used instead of a specific version?

beeme1mr (approved):

src/OpenFeature.Contrib.Providers.ConfigCat/OpenFeature.Contrib.Providers.ConfigCat.csproj
cdonnellytx: .NET considers the single version to be equivalent to be a range. 

The following are equivalent:
```xml
    <PackageReference Include="ConfigCat.Client" Version="9.2.0"/>
    <PackageReference Include="ConfigCat.Client" Version="[9.2.0, )"/>
```

askpt: For reference: https://learn.microsoft.com/en-us/nuget/concepts/package-versioning?tabs=semver20sort#version-ranges 


Notation | Applied rule | Description
-- | -- | --
1.0 | x ≥ 1.0 | Minimum version, inclusive
[1.0,) | x ≥ 1.0 | Minimum version, inclusive

adams85: Yep, as shown in [this table](https://learn.microsoft.com/en-us/nuget/concepts/package-versioning?tabs=semver20sort#version-ranges), the two are equivalent. I would go with the simpler syntax.

beeme1mr: Got it, thanks.

PR #208, opened 2024-05-23
feat: relative weights in fractional, fix injected props

bacherfl:
This PR adds support for relative weights in the fractional evaluator. This new (non breaking) feature has been proposed in https://github.com/open-feature/flagd/issues/1282 (The related PR to implement this in flagd is this one: https://github.com/open-feature/flagd/pull/1313).
In addition to supporting relative weights instead of percentages, the weight value for a distribution item can be omitted. In this case, a default weight of `1` will be applied.

Leaving this in draft until the related PR in flagd has been merged

### DIFF
--- a/src/OpenFeature.Contrib.Providers.Flagd/Resolver/InProcess/CustomEvaluators/FlagdProperties.cs
+++ b/src/OpenFeature.Contrib.Providers.Flagd/Resolver/InProcess/CustomEvaluators/FlagdProperties.cs
@@ -17,7 +17,7 @@ namespace OpenFeature.Contrib.Providers.Flagd.Resolver.InProcess.CustomEvaluator
         internal FlagdProperties(object from)
         {
             //object value;
-            if (from is Dictionary<string, object> dict)
+            if (from is IDictionary<string, object> dict)
             {
                 if (dict.TryGetValue(TargetingKeyKey, out object targetingKeyValue)
                     && targetingKeyValue is string targetingKeyString)
@@ -25,7 +25,7 @@ namespace OpenFeature.Contrib.Providers.Flagd.Resolver.InProcess.CustomEvaluator
                     TargetingKey = targetingKeyString;
                 }
                 if (dict.TryGetValue(FlagdPropertiesKey, out object flagdPropertiesObj)
-                    && flagdPropertiesObj is Dictionary<string, object> flagdProperties)
+                    && flagdPropertiesObj is IDictionary<string, object> flagdProperties)
                 {
                     if (flagdProperties.TryGetValue(FlagKeyKey, out object flagKeyObj)
                         && flagKeyObj is string flagKey)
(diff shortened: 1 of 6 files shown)


toddbaert (approved):

toddbaert:
@aepfli Not sure if .NET is your thing, but since you did a few others, could you quickly skim this?

test/OpenFeature.Contrib.Providers.Flagd.Test/FractionalEvaluatorTest.cs
```diff
@@ -55,7 +55,68 @@ public void Evaluate(string email, string flagKey, string expected)
             // Act & Assert
             var result = evaluator.Apply(rule, data);
             Assert.Equal(expected, result.ToString());
+        }
+
+        [Theory]
+        [MemberData(nameof(FractionalEvaluationTestData.FractionalEvaluationContext), MemberType = typeof(FractionalEvaluationTestData))]
+        public void EvaluateUsingRelativeWeights(string email, string flagKey, string expected)
```
aepfli, line 62: we should evaluate if it makes sense to move the json tests of the java contrib sdk from https://github.com/open-feature/java-sdk-contrib/tree/main/providers/flagd/src/test/resources/fractional to the harness, and use them as test for all our implementation of the fractionals. maybe we are still missing some data, which is part of the test code, but i think this would allow us to be more stable throughout all the packages. Normally there should be a way in each of our contrib sdks to load json files and to parameterize tests in some sort of way (assumption for dotnet)

aepfli (approved):
looks good from my side, i do have a little comment regarding the tests, but this is more like a little nit or improvement for the future

toddbaert:
I found a couple small issues with this when adding some additional e2e tests, working on some fixes.

toddbaert:
@bacherfl I made a couple minor tweaks to the algorithm; we had a slight deviation when no bucking value was supplied. You can see them [here](https://github.com/open-feature/dotnet-sdk-contrib/pull/208/commits/edd04f0c9ac46b0c4764863ea94ac98726a7bb93).

I also added the e2e tests for this new functionality.

Thanks so much!

src/OpenFeature.Contrib.Providers.Flagd/Resolver/InProcess/CustomEvaluators/FlagdProperties.cs
```diff
@@ -17,15 +17,15 @@ internal class FlagdProperties
         internal FlagdProperties(object from)
         {
             //object value;
-            if (from is Dictionary<string, object> dict)
+            if (from is IDictionary<string, object> dict)
```
toddbaert, line 20: This stumped me for a moment. `from` here implements `IDictionary`, but is not actually an instance of `Dictionary`, so this evaluated to false, resulting in no injected props. This fixes it!

PR #212, opened 2024-06-07
fix: refactor OTEL metrics tests to make them more stable

jenshenneberg:
## This PR
- fixes the `MetricsHookTest` by using `MeterProvider.ForceFlush()` rather than waiting for periodic flush to happen

### Notes
The test in question would fail occasionally. This refactor will hopefully make them more stable, as there is the need to wait for an async process to complete has been removed.


askpt (approved):
LGTM! Thanks for this improvement. This change will make the tests more reliable since we are not "waiting" for the meters to be flushed.

beeme1mr (approved):

PR #218, opened 2024-06-19
feat: Change fractional custom op from percentage-based to relative weighting.

aepfli:
for consitency with [golang](https://github.com/open-feature/flagd/issues/1282), [js](https://github.com/open-feature/js-sdk-contrib/issues/946), and java

i am not sure if this is even desired, and I am not that familiar with csharp, so there might be tests missing


aepfli:
i just saw that there is already a pr doing this

Issue #226, opened 2024-07-04
[flagd] Add targeting validation and log warning


toddbaert:
The Go/in-process, Java/in-process, and JS/in-process and standalone versions of flagd log a warning message with invalid targeting rules when they are fetched. We should implement this in the dotnet provider

Targeting rules can be validated with the [flags.json](https://flagd.dev/schema/v0/flags.json) schema in conjunction with the [targeting.json](https://flagd.dev/schema/v0/targeting.json) schema. Both will have to be loaded into a validator library to fully validate. See similar examples in [Go](https://github.com/open-feature/flagd-schemas/blob/8c72c14eebff2a5b20fe2afb90c7ad44c1184ae8/json/flagd_definitions_test.go#L19-L25) and [Java](https://github.com/open-feature/java-sdk-contrib/blob/d468baeb4a54cc712ee95cb1a779c115da766033/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/model/FlagParser.java#L42-L57)

The in-process provider should be modified to use both of these schemas, and validate against them, only logging a warning if validation fails. Only a failure to parse the definition into a [FeatureFlag instance](https://github.com/open-feature/dotnet-sdk-contrib/blob/7cccc8df0de6d9607e045fa62f070f35f20d6a0a/src/OpenFeature.Contrib.Providers.Flagd/Resolver/InProcess/JsonEvaluator.cs#L17) should cause an abortive error.

Logging should be done with the generic `Microsoft.Extensions.Logging.Abstractions` package.

PR #233, opened 2024-07-15
fix: remove Microsoft.Extensions.Logging from flagd provider

### LINKED ISSUE #232: flagd dependencies on Microsoft.Extensions.Logging
I don't think `OpenFeature.Contrib.Providers.Flagd` should depend on `Microsoft.Extensions.Logging`, `Microsoft.Extensions.Logging.Console` or `Microsoft.Extensions.Logging.Debug`.

As well as them just being extra dependencies on the package, the consuming application should really be defining where the logs go, rather than the library.

It could maybe depend on `Microsoft.Extensions.Logging.Abstractions` by taking an `ILogger`, `ILogger<T>` or `ILoggerFactory` in the `InProcessResolver` constructor and passing it through (the OpenFeature SDK itself can take an `ILogger?`), but it could also just be removed entirely for now as it's only used for two small pieces of debug logging.

At the moment
- `StringEvaluator` creates a logger in its constructor but does not use it afterwards.
- `SemVerEvaluator` logs a debug message when catching exceptions from `SemVersion.Parse`, but it could just use `SemVersion.TryParse` instead if it did not do this logging, and avoid the throw and catch. RPC mode does not log anything here.
- `FractionalEvaluator` uses it to debug log "no matching bucket found" but it isn't really necessary to log that and it already gives a reason of `DEFAULT` on the details result. RPC mode does not log anything here.

Here's an example of a commit that would just remove it entirely: https://github.com/open-feature/dotnet-sdk-contrib/commit/8c4875b253abca02f18b52e5f3ff5ccb409da0f8

ghelyar:
## This PR
- Remove `Microsoft.Extensions.Logging` dependencies from `OpenFeature.Contrib.Providers.Flagd`
- Use `SemVersion.TryParse` to avoid a throw and catch

### Related Issues

Fixes #232

### DIFF
--- a/src/OpenFeature.Contrib.Providers.Flagd/OpenFeature.Contrib.Providers.Flagd.csproj
+++ b/src/OpenFeature.Contrib.Providers.Flagd/OpenFeature.Contrib.Providers.Flagd.csproj
@@ -23,9 +23,6 @@
     <!-- The schema.proto file referenced here will be used to automatically generate the Grpc client when executing 'dotnet build' -->
     <!-- The generated files will be placed in ./obj/Debug/netstandard2.0/Protos -->
     <PackageReference Include="JsonLogic.Net" Version="1.1.11" />
-    <PackageReference Include="Microsoft.Extensions.Logging" Version="8.0.0" />
-    <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="8.0.0" />
-    <PackageReference Include="Microsoft.Extensions.Logging.Debug" Version="8.0.0" />
     <PackageReference Include="murmurhash" Version="1.0.3" />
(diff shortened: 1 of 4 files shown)


beeme1mr (approved):

toddbaert:
I'll review this later today.

toddbaert:
@ghelyar nice. Approving, merging and releasing.

I've also updated this issue, which has to do with some (more important) logging: https://github.com/open-feature/dotnet-sdk-contrib/issues/226

toddbaert:
released: https://github.com/open-feature/dotnet-sdk-contrib/pull/235, https://github.com/open-feature/dotnet-sdk-contrib/releases/tag/OpenFeature.Contrib.Providers.Flagd-v0.2.1

PR #309, opened 2025-01-15
feat: Update in-process resolver to support flag metadata #305

### LINKED ISSUE #305: [flagd] Update in-process resolver to support flag metadata
With [this schema change](https://github.com/open-feature/flagd-schemas/pull/173), flag metadata can now be stored in flag configuration files. Specifically, these can contain things like `version` and  flag set `id` information so that consumers can use this data in flag evaluations, caching, telemetry etc.

We need to add this metadata into flags which are evaluated by in-process resolvers. Essentially, we need to take the new properties from the flag definition and include them in metadata returned in the [flag resolution](https://github.com/open-feature/dotnet-sdk-contrib/blob/2f4907eaca791a25cd61000ab81be87f1aa4c764/src/OpenFeature.Contrib.Providers.Flagd/Resolver/InProcess/JsonEvaluator.cs#L211).

Acceptance criteria:

- metadata defined in flag configuration (specifically `id` and `version`) are returned from any flag resolved from this set
- associated tests

chrfwow:
## This PR
Adds support for metadata to the flagd provider

### Related Issues

Fixes #305


src/OpenFeature.Contrib.Providers.Flagd/Resolver/InProcess/JsonEvaluator.cs
```diff
@@ -88,7 +83,59 @@ internal FlagSyncData Parse(string flagConfigurations)
...
+                data.Metadata = new Dictionary<string, object>();
+            }
+            else
+            {
+                foreach (var key in new List<string>(data.Metadata.Keys))
+                {
+                    var value = data.Metadata[key];
+                    if (value is long longValue)
+                    {
+                        data.Metadata[key] = (int)longValue;
```
aepfli, line 99: [question] I don't have read the spec, but why must we transform every long into an int?

src/OpenFeature.Contrib.Providers.Flagd/Resolver/InProcess/JsonEvaluator.cs
```diff
@@ -88,7 +83,59 @@ internal FlagSyncData Parse(string flagConfigurations)
...
+                        continue;
```
aepfli, line 100: [suggestion] Although we know now that based on the current implementation of `VerifyMetaDataValue,` we do have a correct value, what if we extend the method with more checks? we might not check for int, hence i would drop this `continue` and verify all the time
```suggestion
```

src/OpenFeature.Contrib.Providers.Flagd/Resolver/InProcess/JsonEvaluator.cs
```diff
@@ -88,7 +83,59 @@ internal FlagSyncData Parse(string flagConfigurations)
...
+                    continue;
+                }
+
+                foreach (var key in new List<string>(flagConfig.Value.Metadata.Keys))
+                {
+                    var value = flagConfig.Value.Metadata[key];
+                    if (value is long longValue)
+                    {
+                        flagConfig.Value.Metadata[key] = (int)longValue;
+                        continue;
```
aepfli, line 120: [suggestion] Although we know now that based on the current implementation of `VerifyMetaDataValue,` we do have a correct value, what if we extend the method with more checks? we might not check for int, hence i would drop this `continue` and verify all the time
```suggestion
```

src/OpenFeature.Contrib.Providers.Flagd/Resolver/InProcess/JsonEvaluator.cs
```diff
@@ -88,7 +83,59 @@ internal FlagSyncData Parse(string flagConfigurations)
...
+                    }
+
+                    VerifyMetadataValue(key, value);
```
aepfli, line 123: [question] We are now verifying all the time, even if the "global" metadata value gets overwritten by the flag metadata. Should we maybe verify this data once the whole metadata is populated instead of twice?

This would add the benefit of not failing if, e.g. There is faulty metadata in the global, which gets overwritten all the time. wdyt?

aepfli (approved):
Thank you, looks good to me. i just have a few questions/suggestions, but nothing which should block a merge ;)

src/OpenFeature.Contrib.Providers.Flagd/Resolver/InProcess/JsonEvaluator.cs
chrfwow: The `ImmutableMetadata` class only supports int and no long values, but the JSON library only returns longs. So we have to convert every long from the json conversion to an int

src/OpenFeature.Contrib.Providers.Flagd/Resolver/InProcess/JsonEvaluator.cs
chrfwow: I would argue that parsing should fail if the file is incorrect, even if the error would not result in a failure.
Furthermore, this could make problems because we might not know if a global metadata entry will be overwritten in the future. Flags can change at any time, and with them their metadata.

src/OpenFeature.Contrib.Providers.Flagd/Resolver/InProcess/JsonEvaluator.cs
```diff
@@ -88,7 +83,57 @@ internal FlagSyncData Parse(string flagConfigurations)
...
+        }
+
+        private static void VerifyMetadataValue(string key, object value)
+        {
+            if (value is int || value is double || value is string || value is bool)
+            {
+                return;
+            }
+
+            throw new ParseErrorException("Metadata entry for key " + key + " and value " + value +
```
chrfwow, line 135: Is this the proper exception type?

toddbaert: Ya I think it's the best one.

src/OpenFeature.Contrib.Providers.Flagd/Resolver/InProcess/JsonEvaluator.cs
```diff
@@ -88,7 +83,57 @@ internal FlagSyncData Parse(string flagConfigurations)
...
-            return JsonConvert.DeserializeObject<FlagSyncData>(transformed);
+
+            var data = JsonConvert.DeserializeObject<FlagSyncData>(transformed);
+            if (data.Metadata == null)
+            {
+                data.Metadata = new Dictionary<string, object>();
+            }
+            else
+            {
+                foreach (var key in new List<string>(data.Metadata.Keys))
```
toddbaert, line 94: nit: I think it would be more typical to do this functionally with Linq (`.Select`) but that's a nit.

toddbaert (approved):

PR #319, opened 2025-03-29
feat: add OpenFeature REST Evaluation Protocol (OFREP) Provider

weyert:
## This PR
The provider allows users to evaluate feature flags through the OFREP API, making it adaptable to various feature flag management systems that support OFREP specification.

Features:

  - Full implementation of the OpenFeature REST Evaluation Protocol specification
  - Support for all standard flag types (boolean, string, integer, double, object)
  - Advanced caching with ETag support to minimize API calls
  - Configuration options for HTTP client, caching behavior, and authorization

### Related Issues
No related issue 

### Follow-up Tasks
For some reason `dotnet test .` can't see the unit tests make it difficult to fix my unit tests.

### How to test
Manually tested


src/OpenFeature.Providers.Ofrep/Client/IOfrepClient.cs
```diff
@@ -0,0 +1,37 @@
...
+    /// <summary>
+    /// Evaluates all flags for a given context using the OFREP bulk evaluation API.
+    /// </summary>
+    /// <param name="context">The evaluation context.</param>
+    /// <param name="cancellationToken">A token to cancel the operation.</param>
+    /// <returns>The bulk evaluation response containing all evaluated flags.</returns>
+    /// <exception cref="OfrepConfigurationException">Thrown if the request fails due to configuration or network issues and no stale cache is available.</exception>
+    /// <exception cref="JsonException">Thrown if the response cannot be parsed and no stale cache is available.</exception>
+    /// <exception cref="OperationCanceledException">Thrown if the operation is cancelled or times out.</exception>
+    Task<BulkEvaluationResponse> BulkEvaluate(EvaluationContext context, CancellationToken cancellationToken);
```
thomaspoignant, line 36: ✂️

thomaspoignant (changes requested):
Hey,
thanks for your PR, I am not sure to get why we have implemented the endpoints for the client providers, in the context of .NET this is not necessary.

Same with the caching, this is not part of the scope of OFREP.
I'll recommend to read the guidelines for the server providers [here](https://github.com/open-feature/protocol/blob/main/guideline/dynamic-context-provider.md) to remove what is unecessary.

src/OpenFeature.Providers.Ofrep/Client/OfrepClient.cs
```diff
@@ -0,0 +1,670 @@
...
+            throw new ArgumentException("Flag key cannot be null or whitespace", nameof(flagKey));
+        }
+
+        if (string.IsNullOrWhiteSpace(type))
+        {
+            throw new ArgumentException("Type cannot be null or whitespace", nameof(type));
+        }
+#endif
+
+        var cacheKey = $"flag:{flagKey};type:{type};ctx:{(context ?? EvaluationContext.Empty).GenerateETag()}";
```
thomaspoignant, line 121: ✂️

src/OpenFeature.Providers.Ofrep/Client/OfrepClient.cs
```diff
@@ -0,0 +1,670 @@
...
+            this.LogCacheHit(cacheKey);
+            return cachedResponse;
+        }
+
+        this.LogCacheMiss(cacheKey);
+
+        try
+        {
+            var request = this.CreateEvaluationRequest(flagKey, context, cacheKey);
+            var response = await this._httpClient.SendAsync(request, cancellationToken).ConfigureAwait(false);
```
thomaspoignant, line 135: You are not checking all the http error code that can be returned by the evaluation api.

- `400` Bad evaluation request : it will return an error object.
- `401` Unauthorized - You need credentials to access the API
- `403` Forbidden - You are not authorized to access the API
- `404` Flag not found : it will return an error object.
- `429` Rate limit reached on the Flag Management System : When receiving this error code, it is important to read the `Retry-After` header and not call the API before this date.
- `500` Internal server error

src/OpenFeature.Providers.Ofrep/Client/OfrepClient.cs
```diff
@@ -0,0 +1,670 @@
...
+        var etagCacheKey = $"etag:{cacheKey}";
+        if (responseETag != null)
+        {
+            this._cache.Set(etagCacheKey, responseETag, cacheEntryOptions);
+        }
+        else
+        {
+            this._cache.Remove(etagCacheKey);
+        }
+    }
```
thomaspoignant, line 353: ✂️

src/OpenFeature.Providers.Ofrep/Client/OfrepClient.cs
```diff
@@ -0,0 +1,670 @@
...
+    {
+        if (this.TryGetBulkCachedResponse(cacheKey, out var staleResponse) && staleResponse != null)
+        {
+            this.LogBulkStaleCacheReturn(cacheKey, ex.GetType().Name, "BulkEvaluate", ex);
+            return staleResponse;
+        }
+
+        throw new OfrepConfigurationException(
+            $"Failed during OFREP bulk evaluation request to {OfrepPaths.BulkEvaluate}.", ex);
+    }
```
thomaspoignant, line 468: ✂️

src/OpenFeature.Providers.Ofrep/Client/OfrepClient.cs
askpt: This should now be fixed!

src/OpenFeature.Providers.Ofrep/Client/OfrepClient.cs
```diff
@@ -0,0 +1,357 @@
...
+
+        if (!string.IsNullOrEmpty(configuration.AuthorizationHeader))
+        {
+            this._httpClient.DefaultRequestHeaders.Authorization =
+                new AuthenticationHeaderValue("Bearer", configuration.AuthorizationHeader);
+        }
+    }
+
+    /// <inheritdoc/>
+    public async Task<OfrepResponse<T>> EvaluateFlag<T>(string flagKey, string type, T defaultValue,
```
thomaspoignant, line 91: Correct me if I am wrong but it seems that the variable `type` is unused.
Also the type is not part of any request for the flag evaluation API.

src/OpenFeature.Providers.Ofrep/Client/OfrepClient.cs
```diff
@@ -0,0 +1,357 @@
...
+        {
+            throw new ArgumentException("Type cannot be null or whitespace", nameof(type));
+        }
+#endif
+
+        if (this._retryAfterDate.HasValue && this._retryAfterDate.Value > DateTimeOffset.UtcNow)
+        {
+            return new OfrepResponse<T>(flagKey, defaultValue)
+            {
+                ErrorCode = ErrorCodes.ProviderNotReady,
```
thomaspoignant, line 113: In the other server SDK in that case we are returning a `GENERAL` error and not a `ProviderNotReady` error.

See:
- [JS](https://github.com/open-feature/js-sdk-contrib/blob/4dbcbf18703bdb0addbef2f7b3810677433abbfd/libs/providers/ofrep/src/lib/ofrep-provider.ts#L72)
- [GO](https://github.com/open-feature/go-sdk-contrib/blob/b435fb47b3c31d5518079e4f8ed2edf4c8ff23ff/providers/ofrep/internal/evaluate/resolver.go#L68)

thomaspoignant (commented):
I have added a few new remarks about the `type` field and also about the error code returned when rate limited.

toddbaert:
I'll take a look at this tomorrow!

src/OpenFeature.Providers.Ofrep/Configuration/OfrepConfiguration.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// Gets or sets additional HTTP headers to include in requests.
+    /// </summary>
+    [JsonPropertyName("headers")]
+    public Dictionary<string, string>? Headers { get; set; }
+
+    /// <summary>
+    /// Gets or sets the authorization header value.
+    /// </summary>
+    [JsonPropertyName("authorizationHeader")]
+    public string? AuthorizationHeader { get; set; }
```
thomaspoignant, line 32: Do we need those 2 options?
A single headers option is enough if we want to set an authorization header you can do it directly in headers.

This is how it works for both the GO and JS provider.

src/OpenFeature.Providers.Ofrep/Client/OfrepClient.cs
```diff
@@ -0,0 +1,351 @@
...
+            {
+                this._httpClient.DefaultRequestHeaders.TryAddWithoutValidation(header.Key, header.Value);
+            }
+        }
+
+        if (!string.IsNullOrEmpty(configuration.AuthorizationHeader))
+        {
+            this._httpClient.DefaultRequestHeaders.Authorization =
+                new AuthenticationHeaderValue("Bearer", configuration.AuthorizationHeader);
+        }
```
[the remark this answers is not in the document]
askpt: Fixed with https://github.com/open-feature/dotnet-sdk-contrib/pull/319/commits/71d7c07607695fad7fc12b80185c49c80af2ffe7

src/OpenFeature.Providers.Ofrep/Client/OfrepClient.cs
```diff
@@ -0,0 +1,351 @@
...
+        {
+            this._retryAfterDate = DateTimeOffset.UtcNow.Add(retryAfter.Delta.Value);
+        }
+        else if (retryAfter?.Date.HasValue == true)
+        {
+            this._retryAfterDate = retryAfter.Date.Value;
+        }
+        else
+        {
+            this._retryAfterDate = DateTimeOffset.UtcNow.AddMinutes(1);
```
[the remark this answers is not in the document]
askpt: I added this check in case the header is not present but we get `Rate Limit` as a response code. I don't know what is the best approach here, so I decided to add 1 minute window.

src/OpenFeature.Providers.Ofrep/Client/OfrepClient.cs
```diff
@@ -0,0 +1,344 @@
...
+        return new HttpClientHandler
+        {
+            UseProxy = true,
+            AutomaticDecompression = DecompressionMethods.GZip | DecompressionMethods.Deflate
+        };
+    }
+
+    // Define high-performance logging delegates using source generators
+    [LoggerMessage(
+        EventId = 1000,
```
kylejuliandev, line 310: We've picked Id = 1000, what is the significance of 1000 here? Do we have conventions or standards for these?

src/OpenFeature.Providers.Ofrep/Extensions/EvaluationContextExtensions.cs
```diff
@@ -0,0 +1,22 @@
+using OpenFeature.Model;
+
+namespace OpenFeature.Providers.Ofrep.Extensions;
+
+/// <summary>
+/// Extension methods for EvaluationContext class
+/// </summary>
+public static class EvaluationContextExtensions
```
kylejuliandev, line 8: Should this be `public`? I'm not sure anything outside of the assembly would need it; maybe some developers would find it handy? If so maybe it would be better defined in the core sdk itself instead of one of the providers

src/OpenFeature.Providers.Ofrep/Models/OfrepResponse.cs
```diff
@@ -0,0 +1,63 @@
+using System.Text.Json.Serialization;
+
+namespace OpenFeature.Providers.Ofrep.Models;
+
+/// <summary>
+/// Represents a response from the OFREP API for a single flag evaluation.
+/// </summary>
+/// <typeparam name="T">The type of the flag value.</typeparam>
+public class OfrepResponse<T>
```
kylejuliandev, line 9: Should this be `public` too? The OpenFeature `ResolvexxxAsync` core abstraction is what developers will be relying on

Unless they need to access the underlying response object directly?

src/OpenFeature.Providers.Ofrep/Models/OfrepRequest.cs
```diff
@@ -0,0 +1,15 @@
+using System.Text.Json.Serialization;
+
+namespace OpenFeature.Providers.Ofrep.Models;
+
+/// <summary>
+/// Represents a request sent to the OFREP API.
+/// </summary>
+public class OfrepRequest
```
kylejuliandev, line 8: Should this request class be `public` visible?

test/OpenFeature.Providers.Ofrep.Test/Client/OfrepClientTest.cs
```diff
@@ -0,0 +1,1400 @@
...
+        Assert.Contains("Network error", result.ErrorMessage);
+    }
+
+    [Fact]
+    public async Task EvaluateFlag_WithCancellationToken_ShouldRespectCancellation()
+    {
+        // Arrange
+        const string flagKey = "test-flag";
+        const bool defaultValue = false;
+        var cts = new CancellationTokenSource();
```
kylejuliandev, line 168: I think cts is disposable, so you can add a using to dispose of it

Not sure how important this is in a short-lived unit test though. So feel free to ignore

test/OpenFeature.Providers.Ofrep.Test/Client/OfrepClientTest.cs
```diff
@@ -0,0 +1,1400 @@
...
+
+        // Assert
+        Assert.NotNull(result);
+        Assert.True(result.Value);
+        Assert.Single(this._mockHandler.Requests);
+    }
+
+    [Fact]
+    public async Task EvaluateFlag_WithNestedEvaluationContext_ShouldSerializeCorrectly()
+    {
```
kylejuliandev, line 738: Similar to the other evaluation context tests, it would be good to check that the context is serialized as we expect it

test/OpenFeature.Providers.Ofrep.Test/Client/OfrepClientTest.cs
```diff
@@ -0,0 +1,1400 @@
...
+        // Arrange
+        const string flagKey = "concurrent-flag";
+        const bool defaultValue = false;
+        const int concurrentRequests = 10;
+
+        var expectedResponse = new OfrepResponse<bool>(flagKey, true);
+
+        // Setup multiple responses
+        for (int i = 0; i < concurrentRequests; i++)
```
kylejuliandev, line 882: Does the setup need to happen 10 times? would 10 calls to the same method result in the same response anyway?

test/OpenFeature.Providers.Ofrep.Test/Client/OfrepClientTest.cs
```diff
@@ -0,0 +1,1400 @@
...
+        var result = await client.EvaluateFlag(flagKey, defaultValue, EvaluationContext.Empty);
+
+        // Assert
+        Assert.NotNull(result);
+        Assert.Equal("unicode-value", result.Value);
+        Assert.Single(this._mockHandler.Requests);
+
+        // Verify the unicode characters were properly URL encoded
+        var request = this._mockHandler.Requests[0];
+        Assert.Contains("ofrep/v1/evaluate/flags/", request.RequestUri?.ToString());
```
kylejuliandev, line 1249: We don't verify that the uri contains the unicode character in the flag key

src/OpenFeature.Providers.Ofrep/Client/OfrepClient.cs
weyert: No real significance.  I think we could also start with `1` here. I have been searching the web but it doesn't sound like the event id's are shared globally.

src/OpenFeature.Providers.Ofrep/Extensions/EvaluationContextExtensions.cs
weyert: I think we can make it `private`.

src/OpenFeature.Providers.Ofrep/Configuration/OfrepOptions.cs
```diff
@@ -0,0 +1,48 @@
...
+
+/// <summary>
+/// Configuration options for the OFREP provider.
+/// </summary>
+public class OfrepOptions
+{
+    /// <summary>
+    /// Gets or sets the base URL for the OFREP API.
+    /// </summary>
+    [JsonPropertyName("baseUrl")]
```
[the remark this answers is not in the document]
weyert: I have removed them. I can't remember why I added them originally

src/OpenFeature.Providers.Ofrep/Models/OfrepRequest.cs
weyert: Updated

test/OpenFeature.Providers.Ofrep.Test/Client/OfrepClientTest.cs
```diff
@@ -0,0 +1,1400 @@
...
+        // Assert
+        Assert.NotNull(result);
+        Assert.Equal(defaultValue, result.Value);
+        Assert.Equal("parse_error", result.ErrorCode);
+        Assert.Contains("null or empty response", result.ErrorMessage);
+    }
+
+    [Fact]
+    public async Task EvaluateFlag_WithEvaluationContext_ShouldIncludeContextInRequest()
+    {
```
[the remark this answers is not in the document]
weyert: I have updated the tests

src/OpenFeature.Providers.Ofrep/OfrepProvider.cs
```diff
@@ -0,0 +1,186 @@
...
+        CancellationToken cancellationToken)
+    {
+#if NET8_0_OR_GREATER
+        ArgumentNullException.ThrowIfNull(flagKey);
+#else
+        if (flagKey == null)
+        {
+            throw new ArgumentNullException(nameof(flagKey));
+        }
+#endif
```
toddbaert, line 147: If this is just to make use of a language feature, I would say vote to remove the compiler directive and just use the common version. I think adding compiler directives like this takes away from readability so I recommend only doing it when we really have to.

If we "really have to" here for some reason I don't understand, that's OK.

toddbaert (approved):
I don't have as much OFREP expertise as some others, but from an overall provider perspective this looks good to me, at least for a first pass.

test/OpenFeature.Providers.Ofrep.Test/Client/OfrepClientTest.cs
weyert: I have updated this test so it using `using`

.tool-versions
askpt, line 1: Could you please remove this file from the repo? If you want, you can add this file to the .gitignore.

askpt (changes requested):

thomaspoignant:
@weyert Why is this PR closed?

PR #342, opened 2025-04-10
test: Remove FluentAssertions

### LINKED ISSUE #326: Remove FluentAssertions
### Description

FluentAssertions is now a paid library starting with version 8. We should consider transitioning to another library like [shouldly](https://github.com/shouldly/shouldly) or use the assertions instead. 

### More information

- https://github.com/fluentassertions/fluentassertions/pull/2943
- https://xceed.com/products/unit-testing/fluent-assertions/
- https://github.com/open-feature/dotnet-sdk/issues/353

kylejuliandev:
## This PR

- Removes FluentAssertions from dotnet-sdk-contrib repository

### Related Issues

Fixes #326

### Notes

Looks like only the Flipt tests were referencing this so it was relatively straight forward. In some of the tests in the `FlipExtensionTests.cs` file I compare the Value/Structure classes via JSON. I looked at building some kind of Equality comparer but I think this approach might over complicate the solution. I've stuck with using the Xunit base assertions, we could opt to use something like shoudly in the future if readability is a concern.

### DIFF
--- a/test/OpenFeature.Contrib.Providers.Flipt.Test/FlipExtensionsTest.cs
+++ b/test/OpenFeature.Contrib.Providers.Flipt.Test/FlipExtensionsTest.cs
@@ -1,21 +1,28 @@
-using System.Text.Json;
-using FluentAssertions;
 using OpenFeature.Contrib.Providers.Flipt.Converters;
 using OpenFeature.Model;
+using System.Text.Json;
 using Xunit;
+using Xunit.Abstractions;
 
 namespace OpenFeature.Contrib.Providers.Flipt.Test;
 
 public class FlipExtensionsTest
 {
+    private readonly ITestOutputHelper _output;
+
+    public FlipExtensionsTest(ITestOutputHelper output)
+    {
+        _output = output;
+    }
+
     [Fact]
     public void ToStringDictionary_WithEmptyContext_ShouldReturnEmptyDictionary()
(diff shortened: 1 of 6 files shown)


test/OpenFeature.Contrib.Providers.Flipt.Test/FlipExtensionsTest.cs
```diff
@@ -1,21 +1,28 @@
...
-using FluentAssertions;
 using OpenFeature.Contrib.Providers.Flipt.Converters;
 using OpenFeature.Model;
+using System.Text.Json;
 using Xunit;
+using Xunit.Abstractions;
 
 namespace OpenFeature.Contrib.Providers.Flipt.Test;
 
 public class FlipExtensionsTest
```
kylejuliandev, line 9:
```suggestion
public class FliptExtensionsTest
```

askpt:
This PR looks good to me. There are some issues with the formatting. Would you mind running `dotnet format` at the root of this repository to try to clear the check error?

kylejuliandev:
> @kylejuliandev This PR looks good to me. There are some issues with the formatting. Would you mind running `dotnet format` at the root of this repository to try to clear the check error?

Odd! I'll get this sorted later today 😁

askpt (approved):

PR #350, opened 2025-04-19
feat: Add support for basic flags in Feature Management

### LINKED ISSUE #263: Cannot make FeatureManagement provider work with On/Off features declaration
Hi, 

I'm encountering an issue with having the OpenFeature / FeatureManagement provider to work with the On/Off features declaration: 
https://learn.microsoft.com/en-us/azure/azure-app-configuration/feature-management-dotnet-reference?pivots=stable-version#onoff-declaration

featureFlags.json
```
{
    // Define feature flags in config file
    "FeatureManagement": {
        "FeatureT": true, // On feature
    }
}
```

Having this code: 

```
var configurationBuilder2 = new ConfigurationBuilder();
configurationBuilder2.AddJsonFile("featureFlags.json");
IConfiguration featureFlagsConfig = configurationBuilder2.Build();

var featureManagementProvider = new FeatureManagementProvider(featureFlagsConfig);
await OpenFeature.Api.Instance.SetProviderAsync(featureManagementProvider);
var client = OpenFeature.Api.Instance.GetClient();

var feature = await client.GetBooleanValue("FeatureT", false);

```

It always evaluates fo false.

But when provided with this code: 

```
{
  "FeatureManagement": {
    "FeatureT": {
      "Allocation": {
        "DefaultWhenEnabled": "FlagEnabled",
        "DefaultWhenDisabled": "FlagDisabled"
      },
      "Variants": [
        {
          "Name": "FlagEnabled",
          "ConfigurationValue": true
        },
        {
          "Name": "FlagDisabled",
          "ConfigurationValue": false
        }
      ],
      "EnabledFor": [
        {
          "Name": "AlwaysOn"
        }
      ]
    }
  }
}

```

It works fine. 

Basically, we have a lot of boolean flags, which doesn't make sense to introduce that kind of structure per each. 

But perhaps I'm missing something. 

This if just a first step to migrate from FeatureManagement to OpenFeature, before connecting with third party provider. 

Any hints ?

### LINKED ISSUE #349: Feature Management Provider does not support basic flags
I'm new to Open Feature and I have a use case for supporting different providers depending on where the application runs, one of those is Azure App Configuration.

I found the Feature Management Provider but using that failed to return the correct flag values for Feature Flags that didn't have variants and were just basic boolean flags.

I've made some changes and think I have something working, I'm happy to push up a PR for this change

Tazmainiandevil:
## This PR
- Adds support for basic boolean flags in Feature Management Provider

### Related Issues
Fixes #349
Fixes #263 (based on: https://github.com/open-feature/dotnet-sdk-contrib/issues/349#issuecomment-2818462444)


src/OpenFeature.Contrib.Providers.FeatureManagement/FeatureManagementProvider.cs
```diff
@@ -50,6 +50,19 @@ public override async Task<ResolutionDetails<bool>> ResolveBooleanValueAsync(str
...
+            {
+                var exists = false;
+                await foreach (var name in featureManager.GetFeatureNamesAsync().WithCancellation(cancellationToken))
+                {
+                    if (!flagKey.Equals(name, StringComparison.OrdinalIgnoreCase)) continue;
+                    exists = true;
+                    break;
+                }
+                var enabled = await featureManager.IsEnabledAsync(flagKey, context, cancellationToken);
+                return new ResolutionDetails<bool>(flagKey, exists ? enabled : defaultValue);
```
askpt, line 63: This is the general guideline we are following at the moment for flags that don't exist.

```suggestion
                if (exists)
                {
                    return new ResolutionDetails<bool>(flagKey, enabled);
                } else {
                    return new ResolutionDetails<bool>(flagKey, defaultValue, ErrorType.FlagNotFound, Reason.Error);
                }
```

If you don't mind, please update the following lines that do evaluations in this file.

askpt (changes requested):
Thank you so much for your contribution. There is just a comment about the error type that should be specified if the flag is not found. 
@ericpattison What are your thoughts?

src/OpenFeature.Contrib.Providers.FeatureManagement/FeatureManagementProvider.cs
Tazmainiandevil: Thanks for the suggestion, I've updated the code. Are these guidelines in the docs anywhere?

Tazmainiandevil: Thanks, I am also looking to port this Feature Management Provider to Python so knowing those requirements are helpful

askpt (approved):

askpt:
Just waiting now for @ericpattison to look into it 👍

kylejuliandev (approved):

PR #373, opened 2025-04-27
feat: add JSON Schema validation for Flagd provider when in-process mode is used

### LINKED ISSUE #226: [flagd] Add targeting validation and log warning

kylejuliandev:
## This PR

- Adds NJsonSchema to validate Flagd configurations when in-process mode is used. Schemas used are: [Flags](https://flagd.dev/schema/v0/flags.json) and [targeting](https://flagd.dev/schema/v0/targeting.json)

### Related Issues

Fixes #226

### Notes

Not yet quite finished. Needs some unit tests. Initial testing locally seems to suggest it is working as expected

image

I'm not sure how keen I am on extending FlagdConfig.cs with ILogger. An alternative approach might be to make it more first class by tweaking the constructor of the FlagdProvider.cs

```csharp
public FlagdProvider(FlagdConfig config, ILogger logger)
{
    // ... other config setup

    if (logger == null)
    {
        logger = NullLogger.Instance;
    }
    _logger = logger
}
```


src/OpenFeature.Contrib.Providers.Flagd/OpenFeature.Contrib.Providers.Flagd.csproj
```diff
@@ -24,6 +24,7 @@
     <!-- The generated files will be placed in ./obj/Debug/netstandard2.0/Protos -->
     <PackageReference Include="JsonLogic" Version="5.4.0" />
     <PackageReference Include="murmurhash" Version="1.0.3" />
+    <PackageReference Include="NJsonSchema" Version="11.2.0" />
```
askpt, line 27: Will you get a lot of breaking changes if you lower to 11.0.0? Since we are publishing this dependency (NJsonSchema), we shouldn't enforce newest versions.

src/OpenFeature.Contrib.Providers.Flagd/Resolver/InProcess/JsonSchemaValidator.cs
```diff
@@ -0,0 +1,109 @@
...
+                return;
+            }
+
+            if (!flag.IsSuccessStatusCode)
+            {
+                _logger.LogWarning("Unable to retrieve Flagd flags JSON Schema, status code: {StatusCode}", flag.StatusCode);
+                return;
+            }
+
+#if NET5_0_OR_GREATER
```
askpt, line 61: Not as part of this PR, but we should bump the target framework...

test/OpenFeature.Contrib.Providers.Flagd.Test/JsonEvaluatorTest.cs
```diff
@@ -14,8 +15,9 @@ public class UnitTestJsonEvaluator
     public void TestJsonEvaluatorAddFlagConfig()
     {
         var fixture = new Fixture();
+        var jsonSchemaValidator = Substitute.For<IJsonSchemaValidator>();
 
-        var jsonEvaluator = new JsonEvaluator(fixture.Create<string>());
+        var jsonEvaluator = new JsonEvaluator(fixture.Create<string>(), jsonSchemaValidator);
```
askpt, line 20: A lot of copy-paste here. I think we should move into a ctor, and control the flow here if needs any change.

askpt (commented):
Just a minor comment in relation to the NJsonSchema version. The rest are minor nitpicks. Great job!

src/OpenFeature.Contrib.Providers.Flagd/OpenFeature.Contrib.Providers.Flagd.csproj
kylejuliandev: Have bumped it down to 11.0.0, you're right there is no need to tie Flagd to 11.2 if it works just as well at 11.0

test/OpenFeature.Contrib.Providers.Flagd.Test/JsonEvaluatorTest.cs
kylejuliandev: I have refactored these tests now to have the constructor initialise the JsonEvaluator. A lot of duplicated code removed now 😄

askpt (approved):

PR #393, opened 2025-05-05
ci: Add provenance atestations and SBOM

askpt:
## This PR

- adds this new feature

### Related Issues

Fixes #1234523


.github/workflows/release.yml
```diff
@@ -8,7 +8,7 @@ on:
 jobs:
   release-package:
     environment: publish
```
toddbaert, line 10: I think you can remove `environment` here because only the other job actually needs the token

.github/workflows/release.yml
```diff
@@ -8,7 +8,7 @@ on:
...
-    runs-on: windows-latest
+    runs-on: ubuntu-latest
     permissions:
       contents: write # for googleapis/release-please-action to create release commit
       pull-requests: write # for googleapis/release-please-action to create release PR
```
toddbaert, line 14: I think you can remove `packages` here since that's only needed in the next job.

Also, I think you need `issues: write` in the latest version of release please, as we have in Java: https://github.com/open-feature/java-sdk-contrib/blob/main/.github/workflows/release-please.yml#L12

.github/workflows/release.yml
```diff
@@ -22,37 +22,75 @@ jobs:
           token: ${{secrets.GITHUB_TOKEN}}
           default-branch: main
           release-type: simple
+    outputs:
+      release_created: ${{ steps.release.outputs.releases_created }}
+      release_tag_name: ${{ steps.release.outputs.release_tag_name }}
+      paths_released: ${{ steps.release.outputs.paths_released }}
```
toddbaert, line 28: There's a bug in some versions of release please where this is a string, not a bool, so I suggest doing `fromJSON` for all of these, as in Java: https://github.com/open-feature/java-sdk-contrib/blob/f22e9e42f7739d9b59aa97ac37a080d901094e0d/.github/workflows/release-please.yml#L21

toddbaert:
I really hope this weoks, but I recommend you go to a v4 version of `googleapis/release-please-action`.

.github/workflows/release.yml
```diff
@@ -22,37 +22,75 @@ jobs:
           token: ${{secrets.GITHUB_TOKEN}}
```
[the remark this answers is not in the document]
toddbaert: If you do this, you'll also need to add the signoff in the rp config: https://github.com/open-feature/java-sdk-contrib/blob/f22e9e42f7739d9b59aa97ac37a080d901094e0d/release-please-config.json#L3

toddbaert:
> > I really hope this works, but I recommend you go to a v4 version of `googleapis/release-please-action`.
> 
> Thanks @toddbaert for your review! Fixed all of your comments. About v4, I would prefer to do in a separate PR. I also noted that dotnet-sdk is still on v3 also

I'm not sure this will all work in v3. There's some bugs there, especially with the signoff. You can try with v3, but I think we will have problems.

askpt:
> > > I really hope this works, but I recommend you go to a v4 version of `googleapis/release-please-action`.
> > 
> > 
> > Thanks @toddbaert for your review! Fixed all of your comments. About v4, I would prefer to do in a separate PR. I also noted that dotnet-sdk is still on v3 also
> 
> I'm not sure this will all work in v3. There's some bugs there, especially with the signoff. You can try with v3, but I think we will have problems. I'd recommend we just go all the way to v4 and I can help roll back or re-release anything if it breaks. Up to you though.

Okay. I will upgrade. I prefer to have everything on the latest. I will use the same SHA version as java-contrib.

toddbaert:
Let's test this together tomorrow.

PR #399, opened 2025-05-08
ci: Update release workflow to find csproj file dynamically

askpt:
…just SBOM paths

## This PR

This pull request updates the `.github/workflows/release.yml` file to improve the handling of `.csproj` files and simplify the SBOM generation and upload process. The changes ensure more precise file identification and streamline the workflow steps.

**Workflow improvements:**

* Added a step to dynamically find the `.csproj` file in the release directory and set its path as an environment variable (`CSPROJ_PATH`). This replaces the previous wildcard-based approach for specifying the `.csproj` path.
* Updated the SBOM generation step to use the dynamically determined `.csproj` path instead of a hardcoded pattern.

**Simplification of SBOM handling:**

* Changed the SBOM output path from `artifacts/sboms/bom.json` to `bom.json` for simplicity.
* Updated the SBOM upload step to reflect the new simplified path.

### Notes
I found that cycloneDX action doesn't like to resolve paths. I added a script that finds the csproj based on the path so we generate the SBOMs for the released package.
Test run here: https://github.com/open-feature/dotnet-sdk-contrib/actions/runs/14912787947 (I was using this workflow as a test for the path/action creation)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -71,11 +71,16 @@ jobs:
         with:
           subject-path: "${{ matrix.release }}/**/*.nupkg"
 
+      - name: Find csproj file and set as environment variable
+        id: find-csproj
+        run: |
+          csproj_path=$(find "${{ matrix.release }}" -name '*.csproj' -type f -print -quit)
+          echo "CSPROJ_PATH=$csproj_path" >> "$GITHUB_OUTPUT"
+
       - name: Generate JSON SBOM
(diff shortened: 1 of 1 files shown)


toddbaert (approved):

toddbaert:
Too bad the simple paths don't work! But this seems good to me.

toddbaert:
Feel free to make a small change to the env provider docs or something and see if it runs.

PR #410, opened 2025-05-17
feat: Add Dependency Injection extensions for Adding Flagd Provider

### LINKED ISSUE #376: Add fluent DI initialization to flagd

kylejuliandev:
## This PR

_Submitting this as draft for API review and discussion._

Adds extension methods to OpenFeatureBuilder for adding and configuring the FlagdProvider.

### Related Issues

Fixes #376

### Notes

If we're going to add these Extension methods we need to bump the OpenFeature dependency up to >2.2.0. I'm not certain how we want consumers to be using this API to inject the Flagd provider. At the moment I added a delegate approach with builder-like methods for chaining configuration calls. I also added methods that allow consumers to create FlagdProviderOptions and set the property values as they like.

__Delegate approach__

```csharp
builder.Services.AddOpenFeature(config =>
{
    config.AddHostedFeatureLifecycle()
        .AddFlagdProvider(options => options
            .WithHost("localhost")
            .WithPort(8013)
            .WithTls(true))
         .AddFlagdProvider("domain1", options => options
            .WithHost("localhost")
         );
});
```

__Property initialization approach__

```csharp
builder.Services.AddOpenFeature(config =>
{
    config.AddHostedFeatureLifecycle()
        .AddFlagdProvider(new FlagdProviderOptions
            {
                Host = "localhost",
                Port = 8013,
                UseTls = true
            })
        .AddFlagdProvider("domain1", new FlagdProviderOptions
            {
                Host = "localhost"
            })
        );
});
```

I need to finish unit testing and ensuring code coverage is sufficient, but before I proceed I wanted to discuss with others what the API should look like.

### Follow-up Tasks

I have created an integration test to help test that the extension methods are working as expected. However, for some reason `OpenFeature.Hosting` only targets net8 and net9. The tests in dotnet-sdk-contrib target net462, net8, and net9. So I'd have to drop net462 as a tfm to keep this test.

### DIFF
--- a/src/OpenFeature.Contrib.Providers.Flagd/DependencyInjection/FeatureBuilderExtensions.cs
+++ b/src/OpenFeature.Contrib.Providers.Flagd/DependencyInjection/FeatureBuilderExtensions.cs
@@ -1,0 +1,111 @@
+using System;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+using OpenFeature.Contrib.Providers.Flagd;
+
+namespace OpenFeature.DependencyInjection.Providers.Flagd;
+
+/// <summary>
+/// Extension methods for configuring the <see cref="OpenFeatureBuilder"/>.
+/// </summary>
+public static class FeatureBuilderExtensions
+{
+    /// <summary>
+    /// Adds the <see cref="FlagdProvider"/> to the <see cref="OpenFeatureBuilder"/> with default <see cref="FlagdProviderOptions"/> configuration.
+    /// </summary>
+    /// <param name="builder">The <see cref="OpenFeatureBuilder"/> instance to configure.</param>
+    /// <returns>The <see cref="OpenFeatureBuilder"/> instance for chaining.</returns>
+    public static OpenFeatureBuilder AddFlagdProvider(this OpenFeatureBuilder builder)
+        => builder.AddProvider(sp =>
+        {
+            return CreateProvider(sp, null, null);
+        });
+
+    /// <summary>
+    /// Adds the <see cref="FlagdProvider"/> to the <see cref="OpenFeatureBuilder"/> with specified <see cref="FlagdProviderOptions"/> configuration.
+    /// </summary>
+    /// <param name="builder">The <see cref="OpenFeatureBuilder"/> instance to configure.</param>
+    /// <param name="configureOptions">Delegate to configure the <see cref="FlagdProvider"/>.</param>
+    /// <returns>The <see cref="OpenFeatureBuilder"/> instance for chaining.</returns>
+    public static OpenFeatureBuilder AddFlagdProvider(this OpenFeatureBuilder builder, Action<FlagdProviderOptions> configureOptions)
+        => builder.AddProvider(sp =>
+        {
+            return CreateProvider(sp, null, configureOptions);
+        });
+
+    /// <summary>
+    /// Adds the <see cref="FlagdProvider"/> to the <see cref="OpenFeatureBuilder"/> with a specific domain and <see cref="FlagdProviderOptions"/> configuration.
+    /// </summary>
+    /// <param name="builder">The <see cref="OpenFeatureBuilder"/> instance to configure.</param>
+    /// <param name="domain">The unique domain of the provider.</param>
+    /// <param name="configureOptions">Delegate to configure the <see cref="FlagdProvider"/>.</param>
+    /// <returns>The <see cref="OpenFeatureBuilder"/> instance for chaining.</returns>
+    public static OpenFeatureBuilder AddFlagdProvider(this OpenFeatureBuilder builder, string domain, Action<FlagdProviderOptions> configureOptions)
+        => builder.AddProvider(domain, (sp, domain) =>
+        {
+            return CreateProvider(sp, domain, configureOptions);
+        });
+
+    /// <summary>
+    /// Adds the <see cref="FlagdProvider"/> to the <see cref="OpenFeatureBuilder"/> with <see cref="FlagdProviderOptions"/> configuration.
+    /// </summary>
+    /// <param name="builder">The <see cref="OpenFeatureBuilder"/> instance to configure.</param>
+    /// <param name="options">Options to configure <see cref="FlagdProvider"/>.</param>
+    /// <returns>The <see cref="OpenFeatureBuilder"/> instance for chaining.</returns>
+    public static OpenFeatureBuilder AddFlagdProvider(this OpenFeatureBuilder builder, FlagdProviderOptions options)
+        => builder.AddProvider(sp =>
+        {
+            return CreateProvider(sp, null, o => { o = options; });
+        });
+
+    /// <summary>
+    /// Adds the <see cref="FlagdProvider"/> to the <see cref="OpenFeatureBuilder"/> with a specific domain and default <see cref="FlagdProviderOptions"/> configuration.
+    /// </summary>
+    /// <param name="builder">The <see cref="OpenFeatureBuilder"/> instance to configure.</param>
+    /// <param name="domain">The unique domain of the provider.</param>
+    /// <returns>The <see cref="OpenFeatureBuilder"/> instance for chaining.</returns>
+    public static OpenFeatureBuilder AddFlagdProvider(this OpenFeatureBuilder builder, string domain)
+        => builder.AddProvider(domain, (sp, domain) =>
+        {
+            return CreateProvider(sp, domain, null);
+        });
+
+    /// <summary>
+    /// Adds the <see cref="FlagdProvider"/> to the <see cref="OpenFeatureBuilder"/> with a specific domain and <see cref="FlagdProviderOptions"/> configuration.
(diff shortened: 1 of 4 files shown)


askpt:
I prefer the property initialisation approach, especially since we are configuring and not adding any extra service.

src/OpenFeature.Contrib.Providers.Flagd/OpenFeature.Contrib.Providers.Flagd.csproj
```diff
@@ -36,8 +36,8 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
-    <PackageReference Include="Microsoft.CSharp" Version="4.7.0"
-      Condition="'$(TargetFramework)' == 'netstandard2.0'" />
-    <PackageReference Include="OpenFeature" Version="[2.0,3.0)" />
+    <PackageReference Include="Microsoft.CSharp" Version="4.7.0" Condition="'$(TargetFramework)' == 'netstandard2.0'" />
+    <PackageReference Include="OpenFeature" Version="[2.2,3.0)" />
+    <PackageReference Include="OpenFeature.DependencyInjection" Version="[2.2,3.0)" />
```
askpt, line 41: I am inclined to suggest moving this to an `OpenFeatureVersionVariable` since these need to be always in the same version.

test/OpenFeature.Contrib.Providers.Flagd.Test/mykey.pem
askpt, line 1: Should this file be committed to the repository?

askpt (commented):
LGTM. My only concern is about the private key that is committed there. I wonder if we should add it there.

src/OpenFeature.Contrib.Providers.Flagd/OpenFeature.Contrib.Providers.Flagd.csproj
kylejuliandev: Good point, I can add a variable so they are tracked together

test/OpenFeature.Contrib.Providers.Flagd.Test/mykey.pem
kylejuliandev: Probably not, I think `mycert.pem` is only necessary. `mykey.pem` was used to generate the certificate.

`mycert.pem` is necessary because constructing the FlagdProvider initializes the InProcessResolver, which in turn creates a client. During this setup, and only in .NET 462, it will attempt to find a real certificate and load it in.

I can remove `mykey.pem` though if it's just clutter. Ideally I'd like to have the test generate these but I'd need to bring in another package reference so it can be done on .NET 462

src/OpenFeature.Contrib.Providers.Flagd/OpenFeature.Contrib.Providers.Flagd.csproj
```diff
@@ -37,7 +37,10 @@
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
     <PackageReference Include="Microsoft.CSharp" Version="4.7.0" Condition="'$(TargetFramework)' == 'netstandard2.0'" />
-    <PackageReference Include="OpenFeature" Version="[2.2,3.0)" />
-    <PackageReference Include="OpenFeature.DependencyInjection" Version="[2.2,3.0)" />
+    <PackageReference Include="OpenFeature" Version="$(OpenFeatureVersion)" />
+    <PackageReference Include="OpenFeature.DependencyInjection" Version="$(OpenFeatureVersion)" />
   </ItemGroup>
+  <PropertyGroup>
+    <OpenFeatureVersion>[2.2,3.0)</OpenFeatureVersion>
```
chrfwow, line 44: Just keep in mind that this also allows `3.0.0-alpha` and similar versions. If we want to exclude 3.0 previews, we need something like [2.2,2.99999]

arttonoyan:
Hi all 👋

@kylejuliandev Thanks for the great work on this! I’d like to suggest as enhancement to improve the registration and configuration logic for the `FlagdProvider`.

Right now, the provider is registered with manual option mapping and provider creation logic. I believe we can improve this by leveraging the .NET Options pattern alongside a dedicated builder extension method (e.g., `ToFlagdConfigBuilder`). This would align the implementation more closely with idiomatic .NET practices.

### Advantages of this approach:

* **Testability**: Configuration can be easily overridden in unit or integration tests using DI, without hardcoding values.
* **Flexibility**: Supports configuration from multiple sources (code, config files, environment variables) and enables named options for multi-tenant scenarios.
* **Maintainability**: Centralizes config mapping in one place (e.g., `ToFlagdConfigBuilder()`), simplifying updates when new fields are introduced.
* **Cleaner Registration**: Keeps the provider registration concise and consistent across overloads.
* **Familiar to .NET developers**: Follows common patterns, which helps new contributors understand and extend the SDK more easily.

### Example (proposed):

```csharp
public static OpenFeatureBuilder AddFlagdProvider(this OpenFeatureBuilder builder)
{
    builder.Services.AddOptions<FlagdProviderOptions>();
    return builder.AddProvider(sp => CreateProvider(sp, null));
}

public static OpenFeatureBuilder AddFlagdProvider(this OpenFeatureBuilder builder, Action<FlagdProviderOptions> configureOptions)
{
    builder.Services.Configure<FlagdProviderOptions>(configureOptions);
    return builder.AddProvider(sp => CreateProvider(sp, null));
}

public static OpenFeatureBuilder AddFlagdProvider(this OpenFeatureBuilder builder, string domain)
{
    builder.Services.AddOptions<FlagdProviderOptions>(domain);
    return builder.AddProvider(domain, (sp, d) => CreateProvider(sp, d));
}

public static OpenFeatureBuilder AddFlagdProvider(this OpenFeatureBuilder builder, string domain, Action<FlagdProviderOptions> configureOptions)
{
    builder.Services.Configure<FlagdProviderOptions>(domain, configureOptions);
    return builder.AddProvider(domain, (sp, d) => CreateProvider(sp, d));
}

private static FlagdProvider CreateProvider(IServiceProvider provider, string domain)
{
    var logger = provider.GetService<ILogger<FlagdProvider>>() ?? NullLogger<FlagdProvider>.Instance;
    var optionsMonitor = provider.GetRequiredService<IOptionsMonitor<FlagdProviderOptions>>();
    var options = optionsMonitor.Get(domain ?? Options.DefaultName);

    var configBuilder = options.ToFlagdConfigBuilder();
    configBuilder.WithLogger(logger);

    return new FlagdProvider(configBuilder.Build());
}
```

Let me know what you think!

src/OpenFeature.Contrib.Providers.Flagd/DependencyInjection/FeatureBuilderExtensions.cs
```diff
@@ -0,0 +1,85 @@
...
+
+        FlagdProviderOptions options;
+        if (string.IsNullOrEmpty(domain))
+        {
+            options = optionsMonitor.Get(FlagdProviderOptions.DefaultName);
+        }
+        else
+        {
+            options = optionsMonitor.Get(domain);
+        }
```
arttonoyan, line 78: One small suggestion: consider replacing `string.IsNullOrEmpty(domain)` with `string.IsNullOrWhiteSpace(domain)`. 
Also, you can simplify this code by replacing the if/else block with a single line using the ternary operator.

So it would become:
```csharp
// Select options based on domain
var options = string.IsNullOrWhiteSpace(domain)
    ? optionsMonitor.Get(FlagdProviderOptions.DefaultName)
    : optionsMonitor.Get(domain);
```

src/OpenFeature.Contrib.Providers.Flagd/DependencyInjection/FlagdProviderOptions.cs
```diff
@@ -0,0 +1,64 @@
...
+    /// The Port property of the config. Defaults to 8013.
+    /// </summary>
+    public int Port { get; set; } = 8013;
+
+    /// <summary>
+    /// Use TLS for communication between the provider and the host. Defaults to false.
+    /// </summary>
+    public bool UseTls { get; set; } = false;
+
+    /// <summary>
```
arttonoyan, line 30: The summary for `CacheEnabled` is currently empty. Please add a brief description to clarify its purpose.

src/OpenFeature.Contrib.Providers.Flagd/DependencyInjection/FlagdProviderOptions.cs
```diff
@@ -0,0 +1,64 @@
...
+    /// 
+    /// </summary>
+    public bool CacheEnabled { get; set; } = false;
```
arttonoyan, line 33: The summary for `CacheEnabled` is currently empty. Please add a brief description to clarify its purpose.

test/OpenFeature.Contrib.Providers.Flagd.Test/mykey.pem
toddbaert: I think committing tests keys/certs is fine, but if we can avoid the key it's ideal, just for appearences.

askpt (approved):

PR #468, opened 2025-09-24
feat: Implement OFREP default evaluation response

### LINKED ISSUE #462: Implement code default flag support in OFREP client
This issue tracks the implementation of the code default flag support, as introduced in [open-feature/protocol#49](https://github.com/open-feature/protocol/pull/49), into the ofrep client within this repository.
The ADR can be seen here: [0006-optional-value-for-code-defaults.md](https://github.com/open-feature/protocol/blob/main/service/adrs/0006-optional-value-for-code-defaults.md)

## Background
- The referenced PR adds support for code default flags to the protocol, allowing feature flag clients to utilize default values defined in code when no flag value is returned by a provider.

## Tasks

- [ ] - Update the ofrep client to handle code default flags according to the protocol specification.
- [ ] - Add unit/integration tests to validate the new behavior.
- [ ] - Update documentation (README, usage guides) to reflect the support for code default flags.

askpt:
Signed-off-by: André Silva <2493377+askpt@users.noreply.github.com>

## This PR

- adds this new feature

### Related Issues

Fixes #462

### DIFF
--- a/src/OpenFeature.Providers.Ofrep/Models/OfrepEvaluationResponse.cs
+++ b/src/OpenFeature.Providers.Ofrep/Models/OfrepEvaluationResponse.cs
@@ -1,0 +1,31 @@
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace OpenFeature.Providers.Ofrep.Models;
+
+/// <summary>
+/// Represents a raw OFREP evaluation response where the value field may be omitted when deferring to code defaults.
+/// </summary>
+internal sealed class OfrepEvaluationResponse
+{
+    [JsonPropertyName("key")]
+    public string? Key { get; set; }
+
+    [JsonPropertyName("errorCode")]
+    public string? ErrorCode { get; set; }
+
+    [JsonPropertyName("errorDetails")]
+    public string? ErrorMessage { get; set; }
+
+    [JsonPropertyName("value")]
+    public JsonElement Value { get; set; }
+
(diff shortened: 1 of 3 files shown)


src/OpenFeature.Providers.Ofrep/Models/OfrepEvaluationResponse.cs
```diff
@@ -0,0 +1,31 @@
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace OpenFeature.Providers.Ofrep.Models;
+
+/// <summary>
+/// Represents a raw OFREP evaluation response where the value field may be omitted when deferring to code defaults.
+/// </summary>
+internal sealed class OfrepEvaluationResponse
```
kylejuliandev, line 9: How does this differ from `OfrepResponse`? I think Value here is `JsonElement` versus `T` in `OfrepResponse`? Couldn't we just use `OfrepResponse<JsonElement>`?

askpt: Great suggestion! Was playing with `T?` but landed with this second class and this makes it way cleaner. Fixed with https://github.com/open-feature/dotnet-sdk-contrib/pull/468/commits/14982ba4c4b0ea44dd6eeec4df9c62436199c9ff

kylejuliandev (approved):

PR #524, opened 2025-12-09
fix: Add MetadataExtensions for converting metadata to primitive types

### LINKED ISSUE #522: [BUG - OFREP] Flag Metadata not working (parsing issues)
### Observed behavior

When I am trying to retrieve some values from Flag Metadata, it seems it is not working correctly. Please check the code examples attached in the steps to reproduce.

Image

In this case, when you try to do a `Get` operation, the result will always be null, even if there are values in the dictionary.

### Expected Behavior

It should be possible to retrieve the values from Flag Metadata.

### Steps to reproduce

- flagd flag definition:
```json
{
    "$schema": "https://flagd.dev/schema/v0/flags.json",
    "flags": {
        "show-welcome-banner": {
            "state": "ENABLED",
            "variants": {
                "on": true,
                "off": false
            },
            "defaultVariant": "off",
            "metadata": {
                "expireDate": "2015-12-01",
                "author": "admin",
                "numberOfViews": 100
            }
        }
    }
}
```
- docker compose:
```yaml
services:
  flagd:
    image: ghcr.io/open-feature/flagd:latest
    volumes:
      - ./flags:/flags
    command: ["start", "--uri", "file:./flags/demo.flagd.json"]
    ports:
      - "8013:8013"
      - "8016:8016"
```

- sample for ofrep
```cs
#:package OpenFeature.Providers.Ofrep@0.1.3
#:package OpenFeature@2.9.0
#:property PublishAot=false
using OpenFeature;
using OpenFeature.Providers.Ofrep;
using OpenFeature.Providers.Ofrep.Configuration;

// Configure the provider
var config = new OfrepOptions("http://localhost:8016");

// Create and register the provider
var provider = new OfrepProvider(config);
await Api.Instance.SetProviderAsync(provider);

// Use feature flags
var client = Api.Instance.GetClient();

// Boolean flag
var boolFlag = await client.GetBooleanDetailsAsync("show-welcome-banner", false);

var date = boolFlag.FlagMetadata.GetString("expireDate");
var author = boolFlag.FlagMetadata.GetString("author");
var numberOfViews = boolFlag.FlagMetadata.GetInt("numberOfViews");
var numberOfViews2 = boolFlag.FlagMetadata.GetString("numberOfViews");

Console.WriteLine($"Boolean Flag: {boolFlag.Value}");
Console.WriteLine($"Expire Date: {date}");
Console.WriteLine($"Author: {author}");
Console.WriteLine($"Number of Views: {numberOfViews}");
Console.WriteLine($"Number of Views 2: {numberOfViews2}");
```

askpt:
## This PR
Adds a new feature that includes `MetadataExtensions` for converting metadata dictionaries containing `JsonElement` values to primitive types. Unit tests validate the functionality of the `ToPrimitiveTypes` method.

### Related Issues

Fixes #522

### DIFF
--- a/src/OpenFeature.Providers.Ofrep/Extensions/MetadataExtensions.cs
+++ b/src/OpenFeature.Providers.Ofrep/Extensions/MetadataExtensions.cs
@@ -35,11 +35,44 @@ internal static class MetadataExtensions
                 JsonValueKind.Number => jsonElement.GetDouble(),
                 JsonValueKind.True => true,
                 JsonValueKind.False => false,
-                _ => value.ToString() ?? string.Empty
+                JsonValueKind.Object => ConvertJsonObject(jsonElement),
+                JsonValueKind.Array => ConvertJsonArray(jsonElement),
+                JsonValueKind.Null => string.Empty,
+                _ => jsonElement.GetRawText()
             };
         }
 
         // If it's already a primitive type, return as-is
         return value;
+    }
+
+    /// <summary>
+    /// Converts a JsonElement object to a Dictionary with primitive values.
+    /// </summary>
+    /// <param name="jsonElement">The JSON element containing the object.</param>
+    /// <returns>A dictionary with string keys and primitive values.</returns>
+    private static Dictionary<string, object> ConvertJsonObject(JsonElement jsonElement)
+    {
+        var result = new Dictionary<string, object>();
+        foreach (var property in jsonElement.EnumerateObject())
+        {
+            result[property.Name] = ExtractPrimitiveValue(property.Value);
+        }
+        return result;
+    }
+
+    /// <summary>
+    /// Converts a JsonElement array to a List with primitive values.
+    /// </summary>
+    /// <param name="jsonElement">The JSON element containing the array.</param>
+    /// <returns>A list with primitive values.</returns>
+    private static List<object> ConvertJsonArray(JsonElement jsonElement)
+    {
+        var result = new List<object>();
+        foreach (var element in jsonElement.EnumerateArray())
+        {
+            result.Add(ExtractPrimitiveValue(element));
+        }
+        return result;
     }
 }
--- a/test/OpenFeature.Providers.Ofrep.Test/Extensions/MetadataExtensionsTest.cs
+++ b/test/OpenFeature.Providers.Ofrep.Test/Extensions/MetadataExtensionsTest.cs
@@ -1,4 +1,5 @@
 using System.Text.Json;
+using OpenFeature.Model;
 using OpenFeature.Providers.Ofrep.Extensions;
 using Xunit;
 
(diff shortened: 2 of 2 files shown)


src/OpenFeature.Providers.Ofrep/Extensions/MetadataExtensions.cs
```diff
@@ -0,0 +1,46 @@
...
+    {
+        if (value is JsonElement jsonElement)
+        {
+            return jsonElement.ValueKind switch
+            {
+                JsonValueKind.String => jsonElement.GetString() ?? string.Empty,
+                JsonValueKind.Number => jsonElement.GetDouble(),
+                JsonValueKind.True => true,
+                JsonValueKind.False => false,
+                _ => value.ToString() ?? string.Empty
```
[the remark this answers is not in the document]
askpt: Fixed.

test/OpenFeature.Providers.Ofrep.Test/Extensions/MetadataExtensionsTest.cs
```diff
@@ -0,0 +1,317 @@
...
+            { "nullKey", json }
+        };
+
+        // Act
+        var result = metadata.ToPrimitiveTypes();
+
+        // Assert
+        Assert.IsType<string>(result["nullKey"]);
+        Assert.Equal(string.Empty, result["nullKey"]);
+    }
```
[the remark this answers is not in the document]
askpt: Fixed.

test/OpenFeature.Providers.Ofrep.Test/Extensions/MetadataExtensionsTest.cs
```diff
@@ -0,0 +1,317 @@
...
+            { "objectKey", json }
+        };
+
+        // Act
+        var result = metadata.ToPrimitiveTypes();
+
+        // Assert
+        Assert.IsType<string>(result["objectKey"]);
+        Assert.Contains("nested", (string)result["objectKey"]);
+    }
```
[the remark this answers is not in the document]
askpt: Fixed.

src/OpenFeature.Providers.Ofrep/Extensions/MetadataExtensions.cs
```diff
@@ -0,0 +1,78 @@
...
+        {
+            return jsonElement.ValueKind switch
+            {
+                JsonValueKind.String => jsonElement.GetString() ?? string.Empty,
+                JsonValueKind.Number => jsonElement.GetDouble(),
+                JsonValueKind.True => true,
+                JsonValueKind.False => false,
+                JsonValueKind.Object => ConvertJsonObject(jsonElement),
+                JsonValueKind.Array => ConvertJsonArray(jsonElement),
+                JsonValueKind.Null => string.Empty,
```
chrfwow: Shouldn't there be a difference between an empty string and a null value?

askpt: I quite agree with you.
The problem is the Immutable Metadata expects an `object` instead of `object?`. See it here: https://github.com/open-feature/dotnet-sdk/blob/main/src/OpenFeature/Model/ImmutableMetadata.cs

Would you think it makes sense to change the sdk?
cc: @open-feature/sdk-dotnet-maintainers @open-feature/technical-steering-committee

kylejuliandev: If nullable warnings were turned off though this should compile? Although then we lose the null handling safety.

askpt: Yes. But it is something we should avoid. In theory you could also do `JsonValueKind.Null => null!;` but it is something I would not agree with since it has the potential for exceptions.

This is a valid point, since we will do a v3 it might be worth visiting back this later on if you agree @kylejuliandev and @chrfwow

PR #547, opened 2026-01-15
refactor(flagd): Remove HTTP call when validating JSON schema

### LINKED ISSUE #474: InProcessResolver JsonSchemaValidator should be able to get the targeting.json and flags.json without calling flagd.dev
# Problem 
Hi, I use the InProcessResolver but actually it don't works because the JsonSchemaValidator try to call https://flagd.dev to fetch targeting.json and flags.json files.  In an environment without access to public internet this behaviour is broken.  
````csharp
  internal JsonSchemaValidator(HttpClient client, ILogger logger)
    {
        if (client == null)
        {
            client = new HttpClient
            {
                BaseAddress = new Uri("https://flagd.dev"),
            };
        }

        _client = client;
        _logger = logger;
    }
````

# Proposals

## Override BaseAdress in configuration
Add extra property into FlagdProviderOptions and FlagdConfig to override BaseAdress of the HttpClient used into the JsonSchemaValidator 

## Use FlagsRetriever and TargetingRetriever
Provide to JsonSchemaValidator two parameters in its constructor to fetch targeting.json, flags.json JsonSchema, this two parameters are provided by the client code and are free implementations. Client could for example get the  schemas from the file system  or generate them from scratch.

### PR DESCRIPTION
**Pull request by kylejuliandev**
## This PR

Refactors the Json Schema Validator in the Flagd provider to use embedded JSON resources instead of relying on HTTP to call out to flagd.dev. I've added a IFlagdJsonSchemaProvider which we may chose to expose to developers in the event they want to provide their own implementations. At the moment this is flagged as internal. I've removed the HTTP code as this was causing issues and this aligns the implementation with how the Java implementation does it.

### Related Issues

Fixes #474

### DIFF
--- a/src/OpenFeature.Contrib.Providers.Flagd/Resolver/InProcess/FlagdJsonSchemaEmbeddedResourceReader.cs
+++ b/src/OpenFeature.Contrib.Providers.Flagd/Resolver/InProcess/FlagdJsonSchemaEmbeddedResourceReader.cs
@@ -12,14 +12,14 @@ internal sealed class FlagdJsonSchemaEmbeddedResourceReader : IFlagdJsonSchemaPr
     const string TargetingJsonResourceName = "OpenFeature.Contrib.Providers.Flagd.Resources.targeting.json";
     const string FlagJsonResourceName = "OpenFeature.Contrib.Providers.Flagd.Resources.flags.json";
 
-    public Task<string> ReadTargetingSchemaAsync(CancellationToken cancellationToken = default)
+    public Task<string> ReadSchemaAsync(FlagdSchema flagdSchema, CancellationToken cancellationToken = default)
     {
-        return this.ReadAsStringAsync(TargetingJsonResourceName, cancellationToken);
-    }
-
-    public Task<string> ReadFlagSchemaAsync(CancellationToken cancellationToken = default)
-    {
-        return this.ReadAsStringAsync(FlagJsonResourceName, cancellationToken);
+        return flagdSchema switch
+        {
+            FlagdSchema.Targeting => this.ReadAsStringAsync(TargetingJsonResourceName, cancellationToken),
+            FlagdSchema.Flags => this.ReadAsStringAsync(FlagJsonResourceName, cancellationToken),
+            _ => throw new ArgumentOutOfRangeException(nameof(flagdSchema), flagdSchema, null)
+        };
     }
 
     private async Task<string> ReadAsStringAsync(string resourceName, CancellationToken cancellationToken = default)
--- a/src/OpenFeature.Contrib.Providers.Flagd/Resolver/InProcess/IFlagdJsonSchemaProvider.cs
+++ b/src/OpenFeature.Contrib.Providers.Flagd/Resolver/InProcess/IFlagdJsonSchemaProvider.cs
@@ -7,7 +7,5 @@ namespace OpenFeature.Contrib.Providers.Flagd.Resolver.InProcess;
 
 internal interface IFlagdJsonSchemaProvider
 {
-    Task<string> ReadTargetingSchemaAsync(CancellationToken cancellationToken = default);
-
-    Task<string> ReadFlagSchemaAsync(CancellationToken cancellationToken = default);
+    Task<string> ReadSchemaAsync(FlagdSchema flagdSchema, CancellationToken cancellationToken = default);
 }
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: src/OpenFeature.Contrib.Providers.Flagd/Resolver/InProcess/FlagdJsonSchemaEmbeddedResourceReader.cs [hunk 1]
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    public Task<string> ReadFlagSchemaAsync(CancellationToken cancellationToken = default)
+    {
+        return this.ReadAsStringAsync(FlagJsonResourceName, cancellationToken);
+    }
+
+    private async Task<string> ReadAsStringAsync(string resourceName, CancellationToken cancellationToken = default)
+    {
+        var assembly = Assembly.GetAssembly(typeof(FlagdJsonSchemaEmbeddedResourceReader));
```
askpt: I don't think this is Native AoT compatible. I would prefer to try reduce the impact so later it is just a matter of a couple of compilation variables.

askpt (changes requested):
I have three major concerns with this PR:
1. NativeAoT with reflection. See comment.
2. enabling nullables in new classes. See comment.
3. Copy of the `flags.json` and `targeting.json`. For this one, please have a look into the submodule called `flagd-schemas`. It includes these files and we can leverage them being already in the repository instead of copy-pasta.

Other than that looks good to me! Thank you!

src/OpenFeature.Contrib.Providers.Flagd/Resolver/InProcess/FlagdJsonSchemaEmbeddedResourceReader.cs
kylejuliandev: Fixed in [cc5e0ec](https://github.com/open-feature/dotnet-sdk-contrib/pull/547/commits/cc5e0ece1ff8cacb04c51c79bdcd3f2083c8f41c)

I believe `typeof(T).Assembly` is NativeAOT friendly, assuming the assembly is available at compile time (which it is)

src/OpenFeature.Contrib.Providers.Flagd/Resolver/InProcess/IFlagdJsonSchemaProvider.cs
```diff
@@ -0,0 +1,13 @@
...
+namespace OpenFeature.Contrib.Providers.Flagd.Resolver.InProcess;
+
+#nullable enable
+
+internal interface IFlagdJsonSchemaProvider
+{
+    Task<string> ReadTargetingSchemaAsync(CancellationToken cancellationToken = default);
+
+    Task<string> ReadFlagSchemaAsync(CancellationToken cancellationToken = default);
+}
```
arttonoyan: Optional suggestion (API-shaping / future-proofing only):

If `IFlagdJsonSchemaProvider` is intended to be an extension point (custom providers, decorators, testing),
you *could* reduce the interface to a single core method and keep the named helpers as default interface methods.
This keeps implementations minimal and helps avoid breaking changes if additional schemas are introduced later.

Example:

```
public enum FlagdSchema
{
    Targeting,
    Flags
}

public interface IFlagdJsonSchemaProvider
{
    Task<string> ReadSchemaAsync(FlagdSchema schema, CancellationToken cancellationToken = default);
}

public static class FlagdJsonSchemaProviderExtensions
{
    public static Task<string> ReadTargetingSchemaAsync(
        this IFlagdJsonSchemaProvider provider,
        CancellationToken cancellationToken = default) =>
        provider.ReadSchemaAsync(FlagdSchema.Targeting, cancellationToken);

    public static Task<string> ReadFlagSchemaAsync(
        this IFlagdJsonSchemaProvider provider,
        CancellationToken cancellationToken = default) =>
        provider.ReadSchemaAsync(FlagdSchema.Flags, cancellationToken);
}
```

This is not a blocker. The current design is fine.
The suggestion is mainly about long-term extensibility and keeping implementations lightweight.

src/OpenFeature.Contrib.Providers.Flagd/Resolver/InProcess/FlagdJsonSchemaEmbeddedResourceReader.cs
```diff
@@ -0,0 +1,47 @@
...
+        return this.ReadAsStringAsync(FlagJsonResourceName, cancellationToken);
+    }
+
+    private async Task<string> ReadAsStringAsync(string resourceName, CancellationToken cancellationToken = default)
+    {
+        var assembly = typeof(FlagdJsonSchemaEmbeddedResourceReader).Assembly;
+        if (assembly == null)
+        {
+            throw new InvalidOperationException($"Unable to locate assembly for {nameof(FlagdJsonSchemaEmbeddedResourceReader)}.");
+        }
```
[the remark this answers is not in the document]
kylejuliandev: Yep you're right! I've fixed this here [9a177de](https://github.com/open-feature/dotnet-sdk-contrib/pull/547/commits/9a177dee5d22db8f97b7cece5827d33343615668)

PR #576, opened 2026-01-25
test(flagd): Add step definition for config.feature tests

kylejuliandev:
## This PR

Add step definitions for the [config.feature](https://github.com/open-feature/flagd-testbed/blob/e1bfc394159b68310d38a8600f53750f8483de83/gherkin/config.feature) tests. At the moment, one of the in-process tests fails as the default port for the in-process resolver is being set to `8013` instead of `8015`. I'll raise a issue. The `sync-port` tests have been ignored for the in-process resolver due to https://github.com/open-feature/dotnet-sdk-contrib/issues/478. Also the following config options have been skipped as they aren't present in the dotnet flagd provider:

- deadlineMs
- fatalStatusCodes
- targetUri
- providerId
- offlineFlagSourcePath
- offlinePollIntervalMs
- streamDeadlineMs
- keepAliveTime
- retryBackoffMs
- retryBackoffMaxMs
- retryGracePeriod

PR #585, opened 2026-02-03
fix(flagd): Default to correct Port when only Resolver is set

kylejuliandev:
## This PR

As part of #576 I'm adding support for the [config.feature](https://github.com/open-feature/flagd-testbed/blob/ecb3d096c82c52424eeb545cf481f2f1d4794d21/gherkin/config.feature#L105-L112) tests. I noticed one failing test which was the following:

```
  @in-process
  Scenario Outline: Default Config In-Process
    Given an option "resolver" of type "ResolverType" with value "in-process"
    When a config was initialized
    Then the option "<option>" of type "<type>" should have the value "<default>"
    Scenarios:
      | option | type    | default |
      | port   | Integer | 8015    |
```

We aren't defaulting the port to 8015 if only the resolver is set. This is a **change in behaviour** but I think it's warranted as it's likely a source of confusion for developers.

I've also used this opportunity to remove an internal constructor on the `FlagdConfig` class and updated all the tests to use the builder pattern - which is the only way devs can configure Flagd. This lays the foundation for addressing #478.

### DIFF
--- a/src/OpenFeature.Contrib.Providers.Flagd/FlagdConfig.cs
+++ b/src/OpenFeature.Contrib.Providers.Flagd/FlagdConfig.cs
@@ -62,7 +62,7 @@ public class FlagdConfig
             .WithHost(uri.Host)
             .WithPort(uri.Port)
             .WithTls(string.Equals(uri.Scheme, "https", StringComparison.OrdinalIgnoreCase))
-            .WithSocketPath(string.Equals(uri.Scheme, "unix", StringComparison.OrdinalIgnoreCase) ? uri.GetComponents(UriComponents.AbsoluteUri & ~UriComponents.Scheme, UriFormat.UriEscaped) : "");
+            .WithSocketPath(string.Equals(uri.Scheme, "unix", StringComparison.OrdinalIgnoreCase) ? uri.GetComponents(UriComponents.AbsoluteUri & ~UriComponents.Scheme, UriFormat.UriEscaped) : string.Empty);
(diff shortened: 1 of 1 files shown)


src/OpenFeature.Contrib.Providers.Flagd/FlagdConfig.cs
```diff
@@ -51,6 +51,20 @@ public static FlagdConfigBuilder Builder()
...
+        if (uri == null)
+        {
+            throw new ArgumentNullException(nameof(uri));
+        }
+
+        return new FlagdConfigBuilder()
+            .WithHost(uri.Host)
+            .WithPort(uri.Port)
+            .WithTls(string.Equals(uri.Scheme, "https", StringComparison.OrdinalIgnoreCase))
+            .WithSocketPath(string.Equals(uri.Scheme, "unix", StringComparison.OrdinalIgnoreCase) ? uri.GetComponents(UriComponents.AbsoluteUri & ~UriComponents.Scheme, UriFormat.UriEscaped) : "");
```
askpt:
```suggestion
            .WithSocketPath(string.Equals(uri.Scheme, "unix", StringComparison.OrdinalIgnoreCase) ? uri.GetComponents(UriComponents.AbsoluteUri & ~UriComponents.Scheme, UriFormat.UriEscaped) : string.Empty);
```

askpt (approved):
LGTM. Just a minor comment.

Issue #601, opened 2026-03-04
[FEATURE] Graceful defaulting to provided code default


toddbaert:
See parent.

PR #605, opened 2026-03-09
feat!: graceful fallback to code default when no default variant resolved

toddbaert:
## This PR
Adding support for handling an empty defaultVariant to fallback to code default (previously was ERROR).

### Related Issues
https://github.com/open-feature/dotnet-sdk-contrib/issues/601


askpt (commented):
Have a question if we should consider whitespaces in this case. Other than that LGTM!

PR #647, opened 2026-05-12
feat(flagd): add FILE resolver type for local flag definition files

### LINKED ISSUE #308: Offline (File) mode support for flagd
Hello, 
I have a question, does the Flagd provider, support offline definitions? 
If not, is this something that may be added in the future?
Here is the specification: https://flagd.dev/reference/specifications/providers/#offline-file-mode

Thank you.

aedeny:
Add support for reading flagd feature flag definitions directly from a local file, with automatic change detection and live reload.

## This PR

Adds a new **FILE** resolver type that reads a flagd flag definition JSON file from disk, validates it against the JSON schema, and evaluates flags locally via `JsonEvaluator` — no flagd connection required.

### New resolver

- **FileBasedResolver**: reads a flagd flag definition JSON file from disk, validates it against the JSON schema, and evaluates flags locally via `JsonEvaluator`. Supports waiting for the file to appear (configurable deadline, default 5 min) and reloading on changes.
- **File watchers behind an `IFileWatcher` abstraction** (under `Resolver/File/`):
  - `FileSystemMtimeWatcher` (**default**): polls the file's modification time and size to detect changes. Lightweight and works across filesystems.
  - `FileSystemHashWatcher` (**opt-in**): polls file content using MurmurHash-128 to detect changes. Designed for environments where mtime is unreliable (e.g., some NFS / container-mounted volumes). Baseline hash is computed synchronously on `Start()` to prevent race conditions. Enable via `UseHashFileChangeDetection`.
  - The OS `FileSystemWatcher` (with debounce) path has been removed, which also simplified shutdown.

### Configuration

Configuration aligns with the [flagd specification](https://flagd.dev):

| Env var | Builder / DI option | Notes |
| --- | --- | --- |
| `FLAGD_RESOLVER=file` | — | selects the FILE resolver |
| `FLAGD_OFFLINE_FLAG_SOURCE_PATH` | `WithOfflineFlagSourcePath()` / `OfflineFlagSourcePath` | path to the flag definition file |
| `FLAGD_OFFLINE_POLL_MS` | `WithOfflinePollIntervalMs()` / `OfflinePollIntervalMs` | polling interval in ms (covers both watcher modes) |
| `FLAGD_DEADLINE_MS` | `WithDeadlineMs()` / `DeadlineMs` | how long to wait for the file to appear, in ms |
| — | `WithUseHashFileChangeDetection()` / `UseHashFileChangeDetection` | opt into hash-based change detection |

Interval options are now expressed as **int milliseconds** to match the spec's `...Ms` naming.

### Error handling

`FileShare.ReadWrite|Delete` for safe concurrent access, schema validation on init, `ProviderError` events on parse failures without crashing, and `ReaderWriterLockSlim` protects the evaluator during reloads.

### Behavior change

> [!IMPORTANT]
> When in-process / FILE flag resolution targets a flag whose targeting evaluates to `null`, the resolver now returns the **default variant's name** instead of `null` (matching the flagd Java implementation). This is a behavior change relevant for release notes.

### Tests

New and updated tests covering `FileBasedResolver` lifecycle and resolution, both watchers' polling behavior, `FlagdConfig` env var parsing, `FlagdProvider` instantiation, and DI option mapping. Full suite passes on net462 / net8.0 / net9.0 / net10.0.

### Documentation

Updated README with the FILE resolver config table entries, usage examples, the new default (mtime polling) vs. opt-in (hash) detection, and interval tuning guidance.

### Related Issues

Fixes #308


aedeny:
Related to #308

kylejuliandev:
Hey @aedeny, thanks for contributing this change! I've noticed the DCO status is failing. Commits need to be signed off with `git commit -s`.

To add your Signed-off-by line to every commit in this branch:
1. In your local branch, run: `git rebase HEAD~1 --signoff`
1. Force push your changes to overwrite the branch: `git push --force-with-lease origin main`

aedeny:
Hey @kylejuliandev, is there an ETA for the review and merge of this PR? Is there anything else that is required from my part?

kylejuliandev:
> Hey @kylejuliandev, is there an ETA for the review and merge of this PR? Is there anything else that is required from my part?

Hey @aedeny! Apologies, I have been away. I will resume my review this week 😄

src/OpenFeature.Providers.Flagd/Resolver/File/FileSystemHashWatcher.cs
```diff
@@ -0,0 +1,220 @@
...
+            _lastModified = File.GetLastWriteTimeUtc(_filePath);
+            _logger?.LogInformation("Initial file read for file '{FilePath}', Hash={LastFileHash}, Size={LastFileSize}, Modified={LastModified}.",
+                _filePath, BitConverter.ToString(hash).Replace("-", ""), _lastSize, _lastModified);
+        }
+
+        _watcherTask = Task.Run(async () => await WatchFileAsync(_cts.Token).ConfigureAwait(false), _cts.Token);
+    }
+
+    public async ValueTask DisposeAsync()
+    {
```
kylejuliandev, line 63: suggestion: is it worth adding a private bool field (e.g. `_disposed`) which we can check for redundant dispose calls?

src/OpenFeature.Providers.Flagd/Resolver/InProcess/FileBasedResolver.cs
```diff
@@ -0,0 +1,408 @@
...
+            else
+            {
+                _fileWatcher.Dispose();
+            }
+            _fileWatcher = null;
+        }
+
+        _debounceTimer?.Dispose();
+        _evaluatorLock.Dispose();
+        _cts.Dispose();
```
kylejuliandev: suggestion: If the timer fires while LoadFlags() is executing (holding the write lock), Shutdown() calling _evaluatorLock.Dispose() will throw SynchronizationLockException. Would it be safer to cancel the CTS first, then dispose the watcher, then dispose the lock (after draining any in-flight callbacks).

src/OpenFeature.Providers.Flagd/FlagdConfig.cs
```diff
@@ -351,13 +435,18 @@ public FlagdConfig Build()
...
+        }
+
         if (this._config.Port == 0)
         {
             var defaultPortForResolver = this._config.ResolverType switch
             {
                 ResolverType.RPC => 8013,
                 ResolverType.IN_PROCESS => 8015,
-                _ => throw new NotImplementedException("ResolverType does not use Ports.")
+                _ => 8013
```
kylejuliandev: thought: as you return early above I think you can leave the NotImplementedException in here

src/OpenFeature.Providers.Flagd/Resolver/InProcess/FileBasedResolver.cs
```diff
@@ -0,0 +1,408 @@
...
+
+    private void CreateFileWatcher()
+    {
+        if (_useHashFileChangeDetection)
+        {
+            _logger?.LogInformation("File watcher for '{FilePath}' is using content hashing for change detection", _filePath);
+
+            var fileWatcher = new FileSystemHashWatcher(_filePath, _logger, _fileChangePollingInterval);
+
+            fileWatcher.FileChanged += (sender, e) =>
```
kylejuliandev: thought: we subscribe to FileChanged, does this included when a file is deleted?

src/OpenFeature.Providers.Flagd/Resolver/InProcess/FileSystemHashWatcher.cs
```diff
@@ -0,0 +1,220 @@
...
+        using (var fs = new FileStream(
+                   _filePath,
+                   FileMode.Open,
+                   FileAccess.Read,
+                   FileShare.ReadWrite | FileShare.Delete,
+                   bufferSize: 4096,
+                   FileOptions.SequentialScan))
+        {
+            var hash = _murmur128.ComputeHash(fs);
+            return (hash, (int)fs.Length);
```
kylejuliandev: thought: does the int cast overflow if the file is too large?

src/OpenFeature.Providers.Flagd/Resolver/File/FileSystemHashWatcher.cs
aedeny: Good call. Added a `private bool _disposed` field and an early-return guard at the top of `DisposeAsync()`. Since `Dispose()` delegates to `DisposeAsync()`, both paths are now idempotent and won't re-cancel/re-dispose `_cts` or `_murmur128` on a second call. Resolved.

src/OpenFeature.Providers.Flagd/Resolver/InProcess/FileSystemHashWatcher.cs
aedeny: You're right, `(int)fs.Length` overflows for files larger than ~2GB, which would also corrupt the size-based change detection. Changed the size to `long` throughout: the tuple return type is now `(byte[] Hash, long Size)`, the `_lastSize` field is `long`, and the cast is removed. Resolved.

src/OpenFeature.Providers.Flagd/Resolver/InProcess/FileBasedResolver.cs
aedeny: Fixed. The old `_debounceTimer?.Dispose()` returned without waiting for an in-flight callback, so `_evaluatorLock.Dispose()` could run while a timer callback still held the write lock. I now use the `Timer.Dispose(WaitHandle)` overload, which signals only after all pending callbacks complete, and wait on it (10s timeout) before disposing the lock. Final ordering matches your suggestion: cancel CTS → dispose watcher → drain timer callbacks → dispose lock. Resolved.

src/OpenFeature.Providers.Flagd/Resolver/InProcess/FileBasedResolver.cs
aedeny: Previously `Deleted` wasn't subscribed, so a deletion was silently ignored. I added an explicit `Deleted` handler that logs a warning and intentionally retains the last-known-good config (matching the hash watcher behavior). Triggering a reload on delete would be wrong since the file is gone; a subsequent re-creation fires `Created`, which reloads. Resolved.

src/OpenFeature.Providers.Flagd/FlagdConfig.cs
aedeny: Agreed. Since `FILE` returns early, the only way to hit `_` is a new unhandled `ResolverType`, and silently defaulting to 8013 would hide that. Replaced the fallback with `throw new NotImplementedException($"No default port defined for resolver type '{this._config.ResolverType}'.")`. Resolved.

aedeny:
> > Hey @kylejuliandev, is there an ETA for the review and merge of this PR? Is there anything else that is required from my part?
> 
> Hey @aedeny! Apologies, I have been away. I will resume my review this week 😄

@kylejuliandev Welcome back and thank you for reviewing! 😄
I just addressed your comments.

src/OpenFeature.Providers.Flagd/Resolver/File/FileBasedResolver.cs
```diff
@@ -0,0 +1,436 @@
...
+    internal static readonly TimeSpan FileSystemWatcherDebounceInterval = TimeSpan.FromMilliseconds(500);
+
+    public event EventHandler<FlagdProviderEvent> ProviderEvent;
+
+    internal FileBasedResolver(ILogger logger, string filePath,
+        IJsonSchemaValidator jsonSchemaValidator,
+        string sourceSelector = "",
+        bool useHashFileChangeDetection = false,
+        TimeSpan? waitForFileReadyInterval = null,
+        TimeSpan? fileChangePollingInterval = null)
```
kylejuliandev, line 39: suggestion: is it worth adding a config option `FLAGD_HASH_FILE_CHANGE_INTERVAL` and `FLAGD_HASH_FILE_READY_INTERVAL` which would allow a developer to configure this? at the moment it will always default to 5minutes (wait for file) and 10s (polling)

aedeny: Done in d9939f9.

The FILE resolver's watcher timings are now configurable instead of hard-coded:

- **Env vars:** `FLAGD_HASH_FILE_CHANGE_INTERVAL_MS` (hash poll interval) and
  `FLAGD_FILE_READY_INTERVAL_MS` (startup file-ready timeout) — both in milliseconds.
- **Builder:** `WithHashFileChangePollingInterval(TimeSpan)` and `WithFileReadyInterval(TimeSpan)`.
- **DI:** `FlagdProviderOptions.HashFileChangePollingInterval` and `.FileReadyInterval`.

A couple of intentional naming/default deviations worth flagging:

- I used a `_MS` suffix and millisecond values to match the flagd ecosystem
  convention for time-based env vars. Happy to switch to seconds if you'd prefer.
- I dropped `HASH_` from the file-ready variable name, since that timeout applies to
  **both** the hash-watcher and the standard `FileSystemWatcher` path, not just the hash mode.
- Lowered the defaults for faster change detection / readiness: hash poll `1m → 5s`,
  startup file-existence poll `10s → 1s`. Let me know if those defaults seem off for your use cases.

Documented in the README under "Tuning the file watcher intervals."

kylejuliandev (approved):

kylejuliandev:
Great addition to the Flagd provider, thanks @aedeny! Possibly a subsequent or follow-up task could be to add support for the File resolver type to the E2E tests

It would be good to have the others @askpt @bacherfl @toddbaert also take a look 👍

askpt (approved):

src/OpenFeature.Providers.Flagd/FlagdConfig.cs
```diff
@@ -38,9 +44,15 @@ public class FlagdConfig
     internal const string EnvVarMaxEventStreamRetries = "FLAGD_MAX_EVENT_STREAM_RETRIES";
     internal const string EnvVarResolverType = "FLAGD_RESOLVER";
     internal const string EnvVarSourceSelector = "FLAGD_SOURCE_SELECTOR";
+    internal const string EnvVarSourceFilePath = "FLAGD_SOURCE_FILE_PATH";
```
toddbaert: **Naming should match the flagd spec.** The cross-language [spec](https://flagd.dev/reference/specifications/providers/#configuration-options) defines:

- `offlineFlagSourcePath` / `FLAGD_OFFLINE_FLAG_SOURCE_PATH`
- `offlinePollIntervalMs` / `FLAGD_OFFLINE_POLL_MS`
- `deadlineMs` / `FLAGD_DEADLINE_MS` (init/connect timeout)

Java and Go both use these; the shared [Gherkin testbed](https://github.com/open-feature/flagd-testbed) asserts against them. Let's rename here so we don't break public API again in 0.8.x when the file-mode e2e tests get wired up.

- `WithSourceFilePath` -> `WithOfflineFlagSourcePath` (`FLAGD_OFFLINE_FLAG_SOURCE_PATH`)
- `WithHashFileChangePollingInterval` -> `WithOfflinePollIntervalMs` (`FLAGD_OFFLINE_POLL_MS`); IMO drop the "hash" qualifier so the same knob covers whichever watcher mode is active
- `WithFileReadyInterval` -> `WithDeadlineMs` (`FLAGD_DEADLINE_MS`); same semantics as RPC/in-process initialization deadline

Same renames in `FlagdProviderOptions`.

src/OpenFeature.Providers.Flagd/Resolver/File/FileBasedResolver.cs
```diff
@@ -0,0 +1,436 @@
...
+        }
+        finally
+        {
+            _evaluatorLock.ExitWriteLock();
+        }
+    }
+
+    private void CreateFileWatcher()
+    {
+        if (_useHashFileChangeDetection)
```
toddbaert, line 307: IIRC we've had a lot of OS-level inconsistency with `FileSystemWatcher`-style APIs in the past (missed events on Linux overlay/NFS/bind-mounted ConfigMaps, exactly the environments this resolver targets); that's the reason the spec keeps file mode to mtime polling only.

Suggestion: make mtime polling the default (driven by `offlinePollIntervalMs`), keep hash-based polling as an opt-in for FS implementations that don't update mtime reliably, and drop the `FileSystemWatcher` path. Hash detection is a real .NET-specific value-add; just IMO shouldn't be the default. WDYT?

src/OpenFeature.Providers.Flagd/Resolver/InProcess/FileBasedResolver.cs
```diff
@@ -0,0 +1,436 @@
...
+using System.IO;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using OpenFeature.Constant;
+using OpenFeature.Error;
+using OpenFeature.Model;
+
+namespace OpenFeature.Providers.Flagd.Resolver.InProcess;
```
toddbaert: Nit: this lives under `Resolver/InProcess/` but isn't part of the in-process resolver. My preference is a separate `Resolver/File/` so the layout reflects three peer resolvers. Only worth doing if it's cheap alongside the renames in the other comment.

src/OpenFeature.Providers.Flagd/Resolver/InProcess/JsonEvaluator.cs
```diff
@@ -339,7 +339,8 @@ private ResolutionDetails<T> ResolveValue<T>(string flagKey, T defaultValue,
 
                 // if variant is null, revert to default
                 reason = Reason.Default;
-                flagConfiguration.Variants.TryGetValue(flagConfiguration.DefaultVariant,
+                variant = flagConfiguration.DefaultVariant;
```
toddbaert, line 342: Nice catch on the `variant` field. Previously when targeting resolved to null we'd fall back to the default variant's value but leave `variant` null even though the resolution genuinely came from a named variant. Matches what Java does; observable behavior change for existing in-process users (variant goes from null to the default variant name) and worth a note in the release entry, but a strict improvement.

toddbaert (changes requested):
@aedeny Amazing work!

Thanks a lot. The only actual blocker I have is [this](https://github.com/open-feature/dotnet-sdk-contrib/pull/647#discussion_r3454650186). I'd rather not merge this with this naming deviation, because if we release, we will have to break again.

Also consider [this](https://github.com/open-feature/dotnet-sdk-contrib/pull/647#discussion_r3454651046). If you are confident this will work reliably, OK - but it's given us a lot of trouble in other languages, due to cross platform issues in filesystems.

src/OpenFeature.Providers.Flagd/FlagdConfig.cs
aedeny: Renamed to match the flagd spec: `WithOfflineFlagSourcePath`/`FLAGD_OFFLINE_FLAG_SOURCE_PATH`, `WithOfflinePollIntervalMs`/`FLAGD_OFFLINE_POLL_MS` (dropped the hash qualifier so it covers both watcher modes), and `WithDeadlineMs`/`FLAGD_DEADLINE_MS`. The interval options are now `int` milliseconds to match the spec's `...Ms` suffix. Same renames applied to `FlagdProviderOptions` and the README. Done in c22673e.

src/OpenFeature.Providers.Flagd/Resolver/File/FileBasedResolver.cs
aedeny: Agreed — mtime polling is now the default (driven by `offlinePollIntervalMs`), hash-based polling is retained as an opt-in via `UseHashFileChangeDetection` for filesystems that don't update mtime reliably, and the OS `FileSystemWatcher` path is removed. This also simplified shutdown (no debounce timer to drain). Done in c22673e.

src/OpenFeature.Providers.Flagd/Resolver/InProcess/FileBasedResolver.cs
aedeny: Moved the file resolver and its watchers into a dedicated `Resolver/File/` folder (namespace `...Resolver.File`) behind an `IFileWatcher` abstraction, so the layout reflects the peer resolvers. Done in c22673e.

toddbaert (approved):

Issue #195, opened 2024-05-02
[Nuget cleanup] Deprecation of renamed packages


thompson-tomo:
Following on from #154 we should be deprecating the old package and transferring the downloads to the latest version. This is to ensure it is clean & organised for consumers.
More information: https://learn.microsoft.com/en-gb/nuget/nuget-org/deprecate-packages#transfer-popularity-to-a-newer-package

## Packages to be deprecated
This is the list of packages that should be marked as deprecated on the NuGet website. Please only mark as deprecated after the child issue for the provider is marked as completed in issue  #154 
- [x] [OpenFeature.Contrib.Hooks.Otel](https://www.nuget.org/packages/OpenFeature.Contrib.Hooks.Otel) <- Will hopefully be moved to the main SDK repo
- [x] [OpenFeature.Contrib.GOFeatureFlag](https://www.nuget.org/packages/OpenFeature.Contrib.GOFeatureFlag)
- [x] [OpenFeature.Contrib.Providers.Flagd](https://www.nuget.org/packages/OpenFeature.Contrib.Providers.Flagd)
- [ ] [OpenFeature.Contrib.Providers.ConfigCat](https://www.nuget.org/packages/OpenFeature.Contrib.Providers.ConfigCat)
- [ ] [OpenFeature.Contrib.Providers.Flagsmith](https://www.nuget.org/packages/OpenFeature.Contrib.Providers.Flagsmith)
- [ ] [OpenFeature.Contrib.Providers.Statsig](https://www.nuget.org/packages/OpenFeature.Contrib.Providers.Statsig)
- [ ] [OpenFeature.Contrib.Providers.Flipt](https://www.nuget.org/packages/OpenFeature.Contrib.Providers.Flipt)
- [ ] [OpenFeature.Contrib.Provider.FeatureManagement](https://www.nuget.org/packages/OpenFeature.Contrib.Provider.FeatureManagement/0.1.0-preview)
- [ ] [OpenFeature.Contrib.Providers.EnvVar](https://www.nuget.org/packages/OpenFeature.Contrib.Providers.EnvVar)

## Ready to be deprecated/unlisted
This is the list of packages that are ready now to be deprecated on the NuGet website.
- [x] [OpenFeature.Contrib.Provider.Statsig](https://www.nuget.org/packages/OpenFeature.Contrib.Provider.Statsig/0.0.3-preview) <- this should be unlisted

Issue #154, opened 2024-02-12
NuGet package names / namespace inconsistencies for OpenFeature providers


julealgon:
I was browsing the existing provider packages produced by this repo and noticed there is an obvious inconsistency in the way they are named:

https://github.com/orgs/open-feature/packages?ecosystem=nuget

<img width="430" alt="image" src="https://github.com/open-feature/dotnet-sdk-contrib/assets/461579/a0e4b6fb-25f1-4b0f-a50f-f9af530d5605">

Here are the formats in use today:
- `OpenFeature.Contrib.Provider.{VendorInfo}`
- `OpenFeature.Contrib.Providers.{VendorInfo}` 
- `OpenFeature.Contrib.{VendorInfo}`

Besides those inconsistencies, there is also the topic of "should the package name have `.Contrib` on it?". OpenTelemetry packages don't, even those that are from the contrib repo. I personally think that's the best naming strategy as it makes it very easy to find any packages and they are all clearly related by the names regardless of the source.

Will these inconsistencies be fixed? What about the `.Contrib` part, have you considered also dropping that from the packages?

toddbaert:
> I was browsing the existing provider packages produced by this repo and noticed there is an obvious inconsistency in the way they are named:

Hey @julealgon I agree the consistency could be improved. Maybe we could do this with some kind of linting/checking in the repo. Unfortunately renaming a package is certainly puts a burden on consumers since it requires a manual migration for continued features/fixes, so it would be up to the maintainers of each contrib package if they want to do it.

> Besides those inconsistencies, there is also the topic of "should the package name have .Contrib on it?". OpenTelemetry packages don't, even those that are from the contrib repo. I personally think that's the best naming strategy as it makes it very easy to find any packages and they are all clearly related by the names regardless of the source.

I don't feel too strongly either way, but I think there's some value in having `Contrib` in the name. The SDK itself (`OpenFeature`) has no dependencies and is owned by the OpenFeature maintainers (it's likely additional "core" packages like this will be published). Artifacts in the contrib repos are stewarded by the OpenFeature maintainers as well, but especially for components that involve integrations with 3rd party tools or vendors, we rely heavily on those who contribute the components to keep them up to date from a feature perspective.

To put it another way, we take responsibility for the packaging, security, and hosting of these contributions, but not the continuing development of their feature sets; I think `Contrib` might be a good indicator of this difference.

[...]

PR #652, opened 2026-05-19
feat(Unleash): add Unleash provider

PSanetra:
## This PR

- Adds a new OpenFeature provider for [Unleash](https://www.getunleash.io/) using the Unleash .NET SDK v6 and OpenFeature SDK v2
- The provider creates and owns its own `DefaultUnleash` client instance (no external `IUnleash` accepted) to properly participate in the OpenFeature initialization lifecycle via Unleash's `ReadyEvent`/`ErrorEvent` callbacks
- Supports file bootstrap for offline usage (`ToggleBootstrapProvider`), skipping HTTP initialization when configured
- Emits `ProviderConfigurationChanged` events when Unleash fires `TogglesUpdatedEvent`
- Maps OpenFeature `EvaluationContext` to Unleash context (`TargetingKey` → `UserId`, well-known keys like `sessionId`, `remoteAddress`, `environment`, `appName`, `currentTime`, and custom properties)
- Boolean evaluation uses `IsEnabled`; string/int/double/structure use `GetVariant` with payload parsing
- Exposes the Unleash variant `payload-type` as flag metadata in `ResolutionDetails`, matching the Java provider pattern

### Related Issues

None

### Notes

- The provider does not accept an external `IUnleash` instance because the Unleash SDK only allows subscribing to lifecycle events (`ReadyEvent`, `ErrorEvent`, `TogglesUpdatedEvent`) during client construction. Without these events, the provider cannot properly signal readiness or configuration changes to the OpenFeature SDK.
- The variant payload `Type` field (e.g., `"string"`, `"number"`, `"json"`, `"csv"`) is exposed as `payload-type` flag metadata for consumers to inspect. The provider does not alter parsing behavior based on the type — it always treats the value as a raw string and attempts parsing based on the OpenFeature evaluation method called (matching the Java provider behavior).
- Tests use file bootstrap


askpt:
@PSanetra Thank you so much for your contribution. Please note the package name should be `OpenFeature.Providers.Unleash`. Would you mind doing the renaming to fix it?

PSanetra:
@askpt I can do that, but it seems like there is already some inconsistency between the provider package names. E.g. the Flipt provider is called `OpenFeature.Contrib.Providers.Flipt` and the Flagd provider is called `OpenFeature.Providers.Flagd`. Is there a reason for the different names?

askpt:
> @askpt I can do that, but it seems like there is already some inconsistency between the provider package names. E.g. the Flipt provider is called `OpenFeature.Contrib.Providers.Flipt` and the Flagd provider is called `OpenFeature.Providers.Flagd`. Is there a reason for the different names?

@PSanetra We decided to go `OpenFeature.Providers.*` have a look at this issue: #154 or #195. I am trying to rename a few to follow the latest pattern but it takes time. The new providers should follow the latest guidelines.

PSanetra:
@askpt Alright, I have renamed the packages and namespaces :slightly_smiling_face:

src/OpenFeature.Providers.Unleash/ContextTransformer.cs
```diff
@@ -0,0 +1,92 @@
...
+        string userId = null;
+        string sessionId = null;
+        string remoteAddress = null;
+        string environment = null;
+        string appName = null;
+        DateTimeOffset? currentTime = null;
+        var properties = new Dictionary<string, string>();
+
+        // Map targeting key to userId if present
+        if (!string.IsNullOrEmpty(context.TargetingKey))
```
askpt:
```suggestion
        if (!string.IsNullOrWhiteSpace(context.TargetingKey))
```

This way "  " will also count as false.

src/OpenFeature.Providers.Unleash/UnleashProvider.cs
```diff
@@ -0,0 +1,213 @@
...
+        }
+
+        var tcs = new TaskCompletionSource<bool>(TaskCreationOptions.RunContinuationsAsynchronously);
+
+        using var registration = cancellationToken.Register(() => tcs.TrySetCanceled(cancellationToken));
+
+        this._unleash = new DefaultUnleash(this._settings, callback =>
+        {
+            callback.ReadyEvent = _ => tcs.TrySetResult(true);
+            callback.ErrorEvent = evt => tcs.TrySetException(
```
askpt: I believe you should raise a `PROVIDER_ERROR`. similar to what you did with the `EmitConfigurationChanged` in this scenario.

src/OpenFeature.Providers.Unleash/ContextTransformer.cs
```diff
@@ -0,0 +1,92 @@
...
+    private const string EnvironmentKey = "environment";
+    private const string AppNameKey = "appName";
+    private const string CurrentTimeKey = "currentTime";
+
+    /// <summary>
+    /// Transforms an OpenFeature EvaluationContext into an Unleash UnleashContext.
+    /// </summary>
+    /// <param name="context">The OpenFeature evaluation context, may be null.</param>
+    /// <returns>A new UnleashContext populated from the evaluation context.</returns>
+    public static UnleashContext Transform(EvaluationContext context)
```
askpt: Did you consider an extension method for the `EvaluationContext`?

This way it would look like `evaluationContext?.ToUnleashContext()`.

src/OpenFeature.Providers.Unleash/UnleashProvider.cs
```diff
@@ -0,0 +1,213 @@
...
+            ProviderName = ProviderMetadata.Name
+        });
+    }
+
+    /// <inheritdoc />
+    public override Task<ResolutionDetails<bool>> ResolveBooleanValueAsync(
+        string flagKey, bool defaultValue, EvaluationContext context = null, CancellationToken cancellationToken = default)
+    {
+        var unleashContext = ContextTransformer.Transform(context);
+        var result = this._unleash.IsEnabled(flagKey, unleashContext, defaultValue);
```
askpt, line 124: I am not familiar with `Unleash` but, does it return a `Evaluation Details`? I am asking because you are assuming all with be `static` evaluations and in case `flag_not_found` for example, we should populate the `ResolutionDetails` correctly.

This applies to all evaluations.

src/OpenFeature.Providers.Unleash/OpenFeature.Providers.Unleash.csproj
askpt, line 1: Would you mind enabling the `nullables` for this project and the tests?

src/OpenFeature.Providers.Unleash/UnleashProvider.cs
```diff
@@ -0,0 +1,213 @@
...
+        {
+            return Task.FromResult(new ResolutionDetails<int>(flagKey, defaultValue, reason: Reason.Default));
+        }
+
+        if (int.TryParse(resolution.Value.PayloadValue, NumberStyles.Any, CultureInfo.InvariantCulture, out var parsed))
+        {
+            return Task.FromResult(new ResolutionDetails<int>(flagKey, parsed, variant: resolution.Value.VariantName, flagMetadata: resolution.Value.GetFlagMetadata()));
+        }
+
+        throw new TypeMismatchException($"Cannot parse variant payload '{resolution.Value.PayloadValue}' as integer for flag '{flagKey}'.");
```
askpt: I don't know if we should `throw` or return a `ResolutionDetails` with status `Error`. The same applies to the other evaluations where you throw an exception.

I remember having discussions about it. Need some help from @open-feature/technical-steering-committee here

askpt (commented):
Did a quick assessment and it seems a very good quality PR. Thanks for your contribution!

Added some small comments.

src/OpenFeature.Providers.Unleash/ContextTransformer.cs
PSanetra: Done

src/OpenFeature.Providers.Unleash/UnleashProvider.cs
PSanetra: Done

src/OpenFeature.Providers.Unleash/ContextTransformer.cs
PSanetra: Done.
I have also passed in the EvaluationContext that was passed initially to InitializeAsync as baseline. Does this make sense? I am not sure which purpose the EvaluationContext in InitializeAsync has otherwise.

src/OpenFeature.Providers.Unleash/OpenFeature.Providers.Unleash.csproj
PSanetra: Done

src/OpenFeature.Providers.Unleash/UnleashProvider.cs
PSanetra: I have changed this to returning ResolutionDetails with `ErrorType.TypeMismatch`

src/OpenFeature.Providers.Unleash/UnleashProvider.cs
PSanetra: If I have not missed something, I think it is not easily possible to distinguish between disabled and non existing flags. In the non-boolean cases I now return `Reason.Disabled` if the `unleash.GetVariant(...)` call returns `Variant.DISABLED_VARIANT`. But I think this is also returned when the flag does not exist.

I think in theory we could also check if the flag is returned from `unleash.ListKnownToggles()`. But I am not sure if it really returns a flag if it is disabled.

PSanetra:
@askpt I think your comments can now be resolved, but please have a second look.

src/OpenFeature.Providers.Unleash/ContextTransformer.cs
askpt: It is important for the client-side SDKs. For example, you want to do a bulk evaluation in the Initialisation. In this case it is less critical, but it supports anyway.

askpt:
@PSanetra Could you please add yourself as a component owner? https://github.com/open-feature/dotnet-sdk-contrib/blob/main/.github/component_owners.yml

Sorry if I am slowly reviewing this PR. The initialisation and lifecycle are quite complex 😄

PSanetra:
No worries, I have added myself :slightly_smiling_face:

src/OpenFeature.Providers.Unleash/UnleashProvider.cs
```diff
@@ -0,0 +1,309 @@
...
+
+    /// <inheritdoc />
+    public override async Task ShutdownAsync(CancellationToken cancellationToken = default)
+    {
+        var clientTask = this.GetClientTask();
+        if (clientTask != null)
+        {
+            var client = await clientTask.ConfigureAwait(false);
+            client?.Dispose();
+        }
```
kylejuliandev: suggestion(non-blocking): should the field `this._clientTcs` be set to null after we have disposed of it? This would prevent subsequent calls to resolve a flag from throwing a Disposed exception, and instead get a ProviderNotReady error

test/OpenFeature.Providers.Unleash.Test/UnleashProviderConstructorTest.cs
```diff
@@ -0,0 +1,129 @@
...
+
+        var provider = new UnleashProvider(settings);
+        var cts = new CancellationTokenSource();
+        cts.Cancel();
+
+        // Should complete without throwing
+        await provider.InitializeAsync(EvaluationContext.Empty, cts.Token);
+
+        // Evaluations still work (client was created, just not ready)
+        var result = await provider.ResolveBooleanValueAsync("flag", true);
```
kylejuliandev, line 57: thought: looking at the method I don't think the CancellationToken is passed to the Unleash task & client, so if a user needed to cancel during flag evaluation this wouldn't be possible?

PSanetra: @kylejuliandev good catch! I have added `WaitAsync(cancellationToken)` to those places where the client is awaited.

src/OpenFeature.Providers.Unleash/UnleashProvider.cs
PSanetra: Done

src/OpenFeature.Providers.Unleash/UnleashProvider.cs
```diff
@@ -0,0 +1,310 @@
...
+    public override Metadata GetMetadata() => ProviderMetadata;
+
+    /// <inheritdoc />
+    /// <remarks>
+    /// When <see cref="UnleashSettings.ToggleBootstrapProvider"/> is configured, the provider
+    /// completes initialization immediately after client construction since bootstrap data is
+    /// loaded synchronously. Otherwise, initialization waits for the first successful HTTP fetch
+    /// (ReadyEvent) or first error (ErrorEvent) or cancellation.
+    /// </remarks>
+    public override Task InitializeAsync(EvaluationContext context, CancellationToken cancellationToken = default)
```
askpt, line 53: I think the `Initialize` can be simplified. The SDK won't call any `Evaluation` until the `Initialize` is finished.

Instead of then having the TCS as a private field in the class, you would have the `IUnleash`. Makes sense? This would simplify a lot the calls and the execution for this provider.

src/OpenFeature.Providers.Unleash/UnleashProvider.cs
```diff
@@ -0,0 +1,310 @@
...
+/// UnleashProvider is the OpenFeature .NET provider implementation for the Unleash feature flag system.
+/// </summary>
+/// <remarks>
+/// This provider creates and owns its own DefaultUnleash client instance rather than accepting
+/// an external IUnleash instance. This is because the Unleash SDK only allows subscribing to
+/// lifecycle events (ReadyEvent, ErrorEvent) during client construction. Without access to these
+/// events, the provider cannot properly implement the OpenFeature initialization lifecycle
+/// (i.e., signaling readiness only after toggles have been fetched).
+/// </remarks>
+public class UnleashProvider : FeatureProvider
```
askpt:
```suggestion
public sealed class UnleashProvider : FeatureProvider
```

askpt (commented):
Added a comment for the `InitializeAsync` method.

src/OpenFeature.Providers.Unleash/UnleashProvider.cs
PSanetra: Done

src/OpenFeature.Providers.Unleash/UnleashProvider.cs
PSanetra: Done

PSanetra:
@askpt @kylejuliandev I have resolved the remaining concerns.

src/OpenFeature.Providers.Unleash/UnleashProvider.cs
```diff
@@ -0,0 +1,257 @@
...
+    internal void EmitConfigurationChanged()
+    {
+        this.EventChannel.Writer.TryWrite(new ProviderEventPayload
+        {
+            Type = ProviderEventTypes.ProviderConfigurationChanged,
+            ProviderName = ProviderMetadata.Name
+        });
+    }
+
+    internal void EmitProviderError(global::Unleash.Events.ErrorEvent evt)
```
askpt: Would you mind remove this namespace declaration and put it in the usings?


```suggestion
    internal void EmitProviderError(ErrorEvent evt)
```

askpt (commented):
Minor nitpick! I am happy with the rest of the provider. Good work 👍

src/OpenFeature.Providers.Unleash/UnleashProvider.cs
askpt: Fixed with: https://github.com/open-feature/dotnet-sdk-contrib/pull/652/commits/c668e85df5a7b82151896bbcb88c03925259e81f

lukas-reining (approved):

askpt (approved):

kylejuliandev (approved):

PR #671, opened 2026-05-25
feat!: DISABLED is a successful evaluation (still defaults)

### LINKED ISSUE #653: feat: flagd provider in-process mode returns reason=DISABLED for disabled flags
## Context

Part of open-feature/flagd#1965. ADR: [docs/architecture-decisions/disabled-flag-evaluation.md](https://github.com/open-feature/flagd/blob/main/docs/architecture-decisions/disabled-flag-evaluation.md)

The flagd provider (`src/OpenFeature.Contrib.Providers.Flagd`) supports in-process evaluation. Today it returns an error for disabled flags. It should return a successful result with `reason=DISABLED` and no value/variant, matching the updated flagd core behavior.

## Changes

- In-process evaluator: when a flag has `state: DISABLED`, return `reason=DISABLED` with value and variant omitted
- Targeting rules should not be evaluated for disabled flags
- Verify behavior against the flagd-testbed scenarios once open-feature/flagd-testbed#375 lands

jonathannorris:
- Disabled flags now resolve successfully with `reason=DISABLED` instead of `errorCode=FLAG_NOT_FOUND`. See the [ADR](https://github.com/open-feature/flagd/blob/main/docs/architecture-decisions/disabled-flag-evaluation.md).
- Breaking, but barely: resolved values are unchanged (the caller-provided default is still surfaced).
- The break is only observable for consumers that inspect `reason` / `errorCode` on disabled-flag evaluations.

### Related Issues

Closes #653
Part of open-feature/flagd#1965

---

> [!WARNING]
> This PR depends on #673 (RPC `flagd-selector` header wiring) merging first; the cross-flagset disabled scenario in the RPC e2e suite will fail until that lands. All other RPC e2e + in-process scenarios pass.

### DIFF
--- a/src/OpenFeature.Providers.Flagd/Resolver/InProcess/JsonEvaluator.cs
+++ b/src/OpenFeature.Providers.Flagd/Resolver/InProcess/JsonEvaluator.cs
@@ -230,8 +230,11 @@ internal class JsonEvaluator
         {
             if ("DISABLED" == flagConfiguration.State)
             {
-                throw new FeatureProviderException(ErrorType.FlagNotFound,
-                    "FLAG_NOT_FOUND: flag '" + flagKey + "' is disabled");
+                return new ResolutionDetails<T>(
+                    flagKey: flagKey,
+                    defaultValue,
+                    reason: Reason.Disabled
+                );
             }
 
             Dictionary<string, object> combinedMetadata = _flagSetMetadata.ToDictionary(
--- a/src/OpenFeature.Providers.Flagd/Resolver/Rpc/RpcResolver.cs
+++ b/src/OpenFeature.Providers.Flagd/Resolver/Rpc/RpcResolver.cs
@@ -87,8 +87,8 @@ internal class RpcResolver : Resolver
                 FlagKey = flagKey
             }).ConfigureAwait(false);
 
-            // Use code default if variant is empty and reason is DEFAULT (no default variant in flag definition)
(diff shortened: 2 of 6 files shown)


jonathannorris:
Note on why we return `defaultValue` rather than `null` for disabled flags, tagging @suthar26 and @toddbaert since this touches the ADR intent.

The Java provider ([open-feature/java-sdk-contrib#1800](https://github.com/open-feature/java-sdk-contrib/pull/1800)) returns `null` for the value field, and the Java SDK substitutes the caller's code default when it sees `null`. That's the cleanest implementation of the ADR's "omit value, SDK uses code default" intent.

We can't do the same in .NET. The provider interface uses non-nullable value types:

```csharp
Task<ResolutionDetails<bool>> ResolveBooleanValueAsync(string key, bool defaultValue, ...)
//                                                                  ^^^^ can't be null
```

`bool`, `int`, and `double` are value types; they always have a value. There's no `Optional<bool>` in the contract. The .NET SDK also doesn't have fallback logic for this case: `defaultValue` is only substituted on exception, not on a successful resolution with `default(T)`. If we returned `default(bool)` (`false`) and the caller passed `true` as their default, they'd get `false` back.

Go has the same constraint (`GenericResolutionDetail[bool]`) and Go providers handle it the same way, returning `defaultValue` when the flag is disabled.

The correct fix would be adding substitution logic to the .NET SDK itself (when `reason=DISABLED`, use `defaultValue`), but that's a separate change. Until then, returning `defaultValue` from the provider achieves the same end result for the caller.

toddbaert (approved):

suthar26 (approved):

JamieSinn (approved):

PR #685, opened 2026-06-14
fix(flagd): apply in-process flag updates after initial sync

### LINKED ISSUE #684: flagd in-process resolver stops applying flag updates after the initial sync
### Observed behavior

When using the flagd provider with `ResolverType.IN_PROCESS`, the provider correctly evaluates flags using the configuration received on its **initial** sync, but it **never applies any subsequent flag-configuration updates** streamed by flagd. Changing a flag (e.g. editing the synced file, or any source change flagd pushes over its sync stream) is not reflected by the provider until the **process is restarted**.

### Expected behavior

The in-process resolver keeps a long-lived `SyncFlags` server-stream open specifically to receive pushed updates. Flag changes should be applied live, without a restart, and a `ProviderConfigurationChanged` event should be emitted.

### Root cause

In `Resolver/InProcess/InProcessResolver.cs`, `Init()` creates the `TaskCompletionSource` that signals initialization completion **without** `RunContinuationsAsynchronously`:

```csharp
var initComplete = new TaskCompletionSource<bool>();
_ = Task.Run(() => HandleEvents(initComplete));
await initComplete.Task.ConfigureAwait(false);
```

`HandleEvents` calls `tcs.TrySetResult(true)` from **inside** the stream-reading loop, upon receiving the first message. Because TCS continuations run synchronously by default, completing the TCS resumes the `Init()` continuation **synchronously on the stream-reading loop thread**. The caller's post-initialization work then runs on that same thread — and if it blocks it (e.g. an ASP.NET host's `app.Run()`), the loop thread is hijacked and never returns to `call.ResponseStream.MoveNext()`. Subsequent `SyncFlags` messages are therefore never read.

This is why the symptom is deterministic and independent of timing, and why a restart "fixes" it (the restart simply re-reads current state once at init).

I confirmed (instrumented logs + `grpcurl` tapping flagd's `:8015` sync stream) that flagd does push the updated configuration on every change — the gap is purely client-side: the loop reaches the `Sync(...)` for the first message but never loops back to the next `MoveNext()`.

### Reproduction

1. Run flagd with a file source (or any source it can push updates for) and the in-process resolver pointed at flagd's sync port (`:8015`).
2. Set the provider on an app whose startup blocks the calling thread (e.g. ASP.NET `app.Run()` after `await Api.Instance.SetProviderAsync(...)`).
3. Evaluate a flag → correct value.
4. Change the flag at the source. flagd's own evaluation API reflects it; the in-process provider does not, indefinitely.

### Proposed fix

Create the TCS with `TaskCreationOptions.RunContinuationsAsynchronously` so completing it does not run the `Init()` continuation on the loop thread:

```csharp
var initComplete = new TaskCompletionSource<bool>(TaskCreationOptions.RunContinuationsAsynchronously);
```

### Environment

- `OpenFeature.Providers.Flagd` 0.7.0 (in-process resolver). Same loop/TCS code is present back through the `OpenFeature.Contrib.Providers.Flagd` 0.5.0 line.
- flagd 0.16.0; .NET 10; reproduced in Docker.

aedeny:
## This PR

Fixes a bug where the **flagd in-process resolver stops applying flag updates after the initial sync**.

The in-process resolver opens a long-lived `SyncFlags` server-stream and processes pushed flag-configuration updates in `HandleEvents`. The `TaskCompletionSource` used to signal `Init()` completion was created without `RunContinuationsAsynchronously`, so completing it from inside the stream loop (on the first received message) ran the `Init()` continuation **synchronously on the stream-reading loop thread**. The caller's post-initialization work (e.g. an ASP.NET host's `app.Run()`) then took over that thread and the loop never returned to `MoveNext()` — so subsequent `SyncFlags` messages were never read and flag updates were never observed after the initial sync until the process restarted.

The fix creates the TCS with `TaskCreationOptions.RunContinuationsAsynchronously` so the `Init()` continuation no longer runs on the loop thread, letting the stream-reading loop keep running and apply live updates.

### Related Issues

Fixes #684

### Notes

- One-line behavioral change in `Resolver/InProcess/InProcessResolver.cs` (`Init()`).
- The same loop/TCS pattern exists back through the `OpenFeature.Contrib.Providers.Flagd` 0.5.0 line, so the bug isn't specific to 0.7.0.
- Verified independently that flagd pushes the updated configuration on every change (instrumented logs + `grpcurl` tapping flagd's `:8015` sync stream); the gap was purely client-side.

### Follow-up Tasks

- Consider a regression test asserting the in-process resolver applies a **second** `SyncFlags` message (i.e. a config update after init) while the caller thread is blocked.

### How to test

1. Run flagd with a file source and the in-process resolver pointed at flagd's sync port (`:8015`).
2. Set the provider on an app whose startup blocks the calling thread (e.g. ASP.NET `app.Run()` after `await Api.Instance.SetProviderAsync(...)`).
3. Evaluate a flag → correct value.
4. Change the flag at the source. **Before:** the in-process provider keeps returning the old value indefinitely (flagd's own eval API reflects the change). **After:** the provider returns the updated value within a sync round-trip, no restart needed.

### DIFF
--- a/src/OpenFeature.Providers.Flagd/Resolver/InProcess/InProcessResolver.cs
+++ b/src/OpenFeature.Providers.Flagd/Resolver/InProcess/InProcessResolver.cs
@@ -58,7 +58,14 @@ internal class InProcessResolver : Resolver
     {
         await _jsonSchemaValidator.InitializeAsync().ConfigureAwait(false);
 
-        var initComplete = new TaskCompletionSource<bool>();
+        // RunContinuationsAsynchronously is required: this TCS is completed from inside
+        // the HandleEvents stream-reading loop (on the first received message). Without
+        // it, completing the TCS runs the Init() continuation synchronously on the loop
+        // thread, so the caller's post-initialization work (e.g. running the host) takes
+        // over that thread and the loop never returns to read subsequent SyncFlags
+        // messages. The result is that in-process flag configuration updates are never
+        // observed after the initial sync (until the process/stream restarts).
+        var initComplete = new TaskCompletionSource<bool>(TaskCreationOptions.RunContinuationsAsynchronously);
         _ = Task.Run(() => HandleEvents(initComplete));
         await initComplete.Task.ConfigureAwait(false);
     }


kylejuliandev (approved):

aedeny:
Friendly ping @askpt. This PR fixes a bug and is a very small change (essentially a one-liner). Would appreciate a review when you have a chance. Thanks!

chrfwow (approved):

aedeny:
Hey @chrfwow, thanks for the review! I see that all checks have passed. Can this be merged?

chrfwow:
@aedeny I don't have write access, so I cannot merge this PR, but for me it is ready to go

aedeny:
Hey @toddbaert, can this be merged?

PR #697, opened 2026-06-21
fix: prevent flaky RpcResolver tests and event stream hot-spin

askpt:
## Why

`RpcResolver` event-stream tests fail intermittently on `windows-latest` (for example [this run](https://github.com/open-feature/dotnet-sdk-contrib/actions/runs/27900850597)) with:

```
Assert.True() Failure - Expected: True, Actual: False
at RpcResolverTests.HandleEvents_WithNoFlags_CallsFlagdProviderEventHandler
```

## Root cause

The mocked gRPC stream signalled end-of-stream **immediately** once its responses were consumed. `RpcResolver.HandleEvents` only increments its retry counter inside `catch (RpcException)`, so when `MoveNext` returns `false` (rather than throwing) the retry bound is never reached and the inner `await MoveNext()` completes synchronously. The loop then reconnects in an unbounded, never-awaiting **hot CPU spin**.

Because the tests never called `Shutdown()` (unlike `InProcessResolverTests`), each test **leaked a busy-spinning background task**. xUnit runs test classes in parallel, so the accumulated spinners saturated the 2-core Windows runner and the 10s waits timed out. The same loop is a latent production bug: a gracefully closed stream would hot-spin in the field.

## Approach

**Production (`RpcResolver.HandleEvents`)**
- Pass the cancellation token to `MoveNext(token)` so the read is cancellable and shutdown is responsive.
- Catch `OperationCanceledException` for graceful shutdown.
- On a graceful end-of-stream (loop exits while not cancelled), route through the existing `HandleErrorEvent()` retry/backoff path (bounded by `MaxEventStreamRetries`) instead of immediately reconnecting in a tight loop.

**Tests (`RpcResolverTests`)**
- `SetupGrpcStream` now emulates a real long-lived stream: after yielding the configured responses it parks on the supplied cancellation token instead of returning `false` instantly, eliminating the in-test hot-spin.
- Each `HandleEvents_*` test shuts the resolver down in a `finally` and sets `WithMaxEventStreamRetries(1)`, mirroring `InProcessResolverTests`, so no background task leaks between tests.
- Blocking `AutoResetEvent.WaitOne(10_000)` waits are replaced with the existing async `Utils.AssertUntilAsync` helper, so tests no longer block a thread-pool thread or depend on real-time scheduling under CPU pressure. All existing assertions are preserved.

## Notes for reviewers

- The behavior change worth a careful look: a graceful end-of-stream now emits a `ProviderError` and backs off (bounded) rather than silently hot-reconnecting. In practice flagd closes the stream via an `RpcException`, so this path is defensive, but it removes the unbounded spin at the source.

## Verification

- Full Flagd suite green on net8.0 (166) and net9.0 (166); `src` builds on all TFMs including net462.
- `RpcResolverTests` stable across 5 consecutive runs (~390ms each); the net8.0 suite dropped from ~46s to under 1s now that the spin is gone.
- `dotnet format --verify-no-changes` clean on the changed files.

### DIFF
--- a/src/OpenFeature.Providers.Flagd/Resolver/Rpc/RpcResolver.cs
+++ b/src/OpenFeature.Providers.Flagd/Resolver/Rpc/RpcResolver.cs
@@ -271,7 +271,7 @@ internal class RpcResolver : Resolver
                 // tight loop.
                 if (!token.IsCancellationRequested)
                 {
-                    await this.HandleErrorEvent().ConfigureAwait(false);
+                    await this.HandleErrorEvent(token).ConfigureAwait(false);
                 }
             }
             catch (OperationCanceledException)
@@ -284,7 +284,7 @@ internal class RpcResolver : Resolver
             }
             catch (RpcException)
             {
-                await this.HandleErrorEvent().ConfigureAwait(false);
+                await this.HandleErrorEvent(token).ConfigureAwait(false);
             }
         }
     }
@@ -327,8 +327,13 @@ internal class RpcResolver : Resolver
         }
     }
 
-    private async Task HandleErrorEvent()
+    private async Task HandleErrorEvent(CancellationToken token)
     {
+        if (token.IsCancellationRequested)
+        {
+            return;
+        }
+
         this._eventStreamRetries++;
 
         if (this._eventStreamRetries > this._config.MaxEventStreamRetries)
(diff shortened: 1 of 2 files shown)


askpt:
@coderabbitai full review

kylejuliandev (approved):

toddbaert (approved):
@askpt perfect! Thank you.

PR #711, opened 2026-07-30
chore(configcat)!: Update ConfigCat provider to use the latest ConfigCat SDK

adams85:
## This PR
Updates ConfigCat providers to use the [latest ConfigCat SDK for .NET](https://github.com/configcat/.net-sdk/releases/tag/v10.0.0).

Also, 
- updates the project to use [nullable reference types](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/null-safety/nullable-reference-types),
- improves the evaluation context to user object mapping to bring the behavior closer to other ConfigCat OpenFeature providers. This comes with some breaking changes though:
  - If `targetingKey` is specified, that will be used as the identifier user attribute, instead of `id` or `identifier`.
  - If `email` or `country` occurs in multiple cases - e.g. `email` and `Email` - the first non-null occurrence will be mapped to the corresponding user attribute, instead of the last occurrence.
  - All context entries except for the ones with the exact key `Identifier`, `Email` and `Country` will also be included in the custom user attributes.
  - Entries with non-string values are now also mapped as custom attributes, rather than string-valued entries only.

### Follow-up Tasks

The PR introduces significant breaking changes, so it should be released with an increased major version.

### DIFF
--- a/src/OpenFeature.Contrib.Providers.ConfigCat/ConfigCatProvider.cs
+++ b/src/OpenFeature.Contrib.Providers.ConfigCat/ConfigCatProvider.cs
@@ -24,7 +24,7 @@ public sealed class ConfigCatProvider : FeatureProvider
     /// <param name="configBuilder">The action used to configure the client.</param>
     /// <exception cref="ArgumentNullException"><paramref name="sdkKey"/> is <see langword="null"/>.</exception>
     /// <exception cref="ArgumentException"><paramref name="sdkKey"/> is an empty string or in an invalid format.</exception>
-    public ConfigCatProvider(string sdkKey, Action<ConfigCatClientOptions> configBuilder = null)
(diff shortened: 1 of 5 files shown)


z4kn4fein (approved):

kylejuliandev (approved):

askpt:
Thanks for your help @adams85! I marked as `!` in the title to report a "breaking change".

@kylejuliandev, before we release, since this is a major breaking change, what do you think if we do a namespace/package naming cleanup as part of #195

adams85:
Thank you for the quick response.

I think this is a good opportunity to change the package/namespace naming. Let me know if we can help with that.

askpt:
> Thank you for the quick response.
> 
> I think this is a good opportunity to change the package/namespace naming. Let me know if we can help with that.

Created #713 to track it. 👍

askpt (approved):

PR #737, opened 2026-09-17
fix(flagsmith): do not send the targeting key as an identity trait

vpetrusevici:
## This PR
`EvaluationContext.AsDictionary()` includes the `targetingKey` entry, so `GetFlags` was passing the targeting key to `GetIdentityFlags` twice: once as the Flagsmith identifier and once more as a trait. The targeting key identifies the identity, it is not an attribute of it, so it now gets filtered out of the trait list.

Also, a few small clean-ups in the same file:

* give `TryParseDelegate` an explicit accessibility modifier
* make the pure helpers `TryParseValue` / `ConvertValue` static
* rewrite `ConvertValue` as a switch so the null/array/object/scalar cases read as one decision
* narrow `TryParseValue`'s catch-all to `JsonException` and reject a null/blank flag value up front instead of relying on `JsonNode.Parse` throwing
* `readonly static` -> `static readonly`
* drop the stale `TargetingKey` property from the `IFlagsmithProviderConfiguration` snippet in the README; it no longer exists on the interface

### Automated

```bash
dotnet test test/OpenFeature.Contrib.Providers.Flagsmith.Test/OpenFeature.Contrib.Providers.Flagsmith.Test.csproj
```

Two tests cover the fix:

- `GetValue_ForEvaluationContextWithTargetingKey_DoesNotSendTargetingKeyAsTrait` (new) — asserts
  that an evaluation context with one trait plus a targeting key reaches
  `IFlagsmithClient.GetIdentityFlags` as the identifier `"233"` and a trait list containing
  exactly `key1`.
- `GetValue_ForEnabledFeatureWithEvaluationContext_ReturnCorrectValue` (tightened) — its trait
  matcher went from `x.Count > 6` to `x.Count == 6`. The loose `> 6` was only ever satisfied
  because `targetingKey` was leaking in as a seventh trait, so the old assertion was passing
  *on* the bug.

To confirm the tests actually catch the regression, revert `FlagsmithProvider.GetFlags` to the
previous implementation (drop the `.Where(...)` on the trait projection) and re-run — both tests
fail.

### Manual

Against a real Flagsmith environment, evaluate a flag with a targeting key set:

```csharp
var context = EvaluationContext.Builder()
    .SetTargetingKey("my-identity-id")
    .Set("my-trait-key", "my-trait-value")
    .Build();

await client.GetBooleanValueAsync("myBoolFlag", false, context);
```

Then open that identity in the Flagsmith dashboard. Before this change it carried an extra
`targetingKey` trait whose value duplicated the identifier; after the change only
`my-trait-key` is present. Existing identities that already picked up the stray trait keep it —
Flagsmith does not remove traits that are no longer sent — so clear it manually if any segment
rule was written against it.


vpetrusevici:
@matthewelwell Hi! Can you check it please? Minor improve

kylejuliandev (approved):

vpetrusevici:
@matthewelwell kind reminder :)

